PR #16, opened 2021-03-17
user trust phase 0

obernardovieira:
```javascript
export interface IManagerDetailsBeneficiary {
    address: string;
    username: string | null;
    timestamp: number;
    claimed: string;
    blocked: boolean;
    // this new fields can be null during this transition phase
    verifiedPN: boolean | undefined;
    suspect: boolean | undefined;
}

export interface IUserHello {
    rates: ExchangeRatesAttributes[];
    isBeneficiary: boolean;
    isManager: boolean;
    community?: ICommunity;
    blocked: boolean;
    // this new fields can be null during this transition phase
    verifiedPN: boolean | undefined;
    suspect: boolean | undefined;
}
```
also, 

* interface `CommunityAttributes` also contains `suspect: UbiCommunitySuspect[];`. When loading community info with `/community/publicid/:publicId` it only returns the most recent suspicious activity if it was less than two days ago.

* `/user/authentication` and `/user/hello` now requires `phone` with the user phone number. For `/user/hello` it's temporary, only to fill users info with require logout/login.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,30 @@
+version: 2.1
+
+orbs:
+  coveralls: coveralls/coveralls@1.0.4
+
+jobs:
+  coverage:
+    docker:
+      - image: circleci/node:14.15.1
+      - image: circleci/postgres:12-alpine
+        environment:
+          POSTGRES_USER: citest
+          POSTGRES_DB: impactmarkettest
+          POSTGRES_PASSWORD: test101
+    steps:
+      - checkout
+      - run: yarn
+      - run:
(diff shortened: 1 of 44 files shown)


obernardovieira:
@Jonathansoufer changes to `types/endpoint.ts` are mentioned above. Further changes might happen.

obernardovieira:
interface `CommunityAttributes` also contains `suspect?: UbiCommunitySuspect[];`. When loading community info with `/community/publicid/:publicId` it only returns the most recent suspicious activity if it was less than two days ago.

PR #40, opened 2021-05-04
Refactor/new endpoints

obernardovieira:


src/api/routes/community.ts
```diff
@@ -6,12 +6,17 @@ import multer from 'multer';
 import { adminAuthentication, authenticateToken } from '../middlewares';
 
 export default (app: Router): void => {
+    const controller = new communityController.CommunityController();
```
Jonathansoufer, line 9: Why is this necessary? Couldn't we just use the communityController directly instead of an instance of it or a better name? I thought is just an extra step to understand once 'controler' naming isn't clear enough because throughout the file you need to find this line to then understand what exactly is the controller.

Jonathansoufer (approved):
Just left one comment. Overall, looks great.

src/api/routes/community.ts
obernardovieira: Forget it for now. It's a huge mess because of the migration.
I understand what you mean. In this case, I'm trying a different approach. Instead of importing an entire file and use functions (as before), I'm trying to create an instance of a class. I'm curious to see what works best and is easier/better for development.

PR #84, opened 2021-08-03
[IPCT1-179] user should be able to migrate account asscociated with phone number

dev-jotape:
This PR fixes [IPCT1-179] at https://impactmarket.atlassian.net/browse/IPCT1-179

## Changes
- change authentication service to validate if the user already has another account and enable account migration.

## New
- included column "active" on user table.
- included "overwrite" as input parameter in endpoint /auth

## Tests
- added some integration tests on user.test.ts

### DIFF
--- a/src/interfaces/app/user.ts
+++ b/src/interfaces/app/user.ts
@@ -103,5 +103,4 @@ export interface UserCreationAttributes {
     pushNotificationToken?: string;
 
     trust?: AppUserTrustCreation;
-    overwrite?: boolean;
 }
--- a/src/interfaces/app/user.ts
+++ b/src/interfaces/app/user.ts
@@ -78,6 +78,7 @@ export interface User {
     children: number | null;
     lastLogin: Date;
     suspect: boolean;
+    active: boolean;
 
     // timestamps
     createdAt: Date;
@@ -102,4 +103,5 @@ export interface UserCreationAttributes {
     pushNotificationToken?: string;
 
     trust?: AppUserTrustCreation;
+    overwrite?: boolean;
 }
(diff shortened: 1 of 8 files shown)


src/interfaces/app/user.ts
```diff
@@ -102,4 +103,5 @@ export interface UserCreationAttributes {
     pushNotificationToken?: string;
 
     trust?: AppUserTrustCreation;
+    overwrite?: boolean;
```
obernardovieira, line 106: this interface only contains model-related variables. `overwrite` isn't part of the User model or any model associated with it.
I see you've used this because of `authenticate`. As an alternative, I suggest changing `authenticate` to `authenticate(user: UserCreationAttributes & { overwrite?: boolean })` or `authenticate(user: UserCreationAttributes, overwrite: boolean =  false)` (I prefer the latter).

tests/integration/services/user.test.ts
```diff
@@ -182,6 +183,107 @@ describe('user service', () => {
...
+                trust: {
+                    phone,
+                },
+                overwrite: true,
+            });
+
+            const findUser = await sequelize.models.UserModel.findOne({
+                where: { address: firstAddress },
+            });
+            const user = findUser?.toJSON() as UserModel
```
obernardovieira, line 238:
```suggestion
            const user = findUser?.toJSON() as User
```

tests/integration/services/user.test.ts
```diff
@@ -182,6 +183,107 @@ describe('user service', () => {
...
+
+            const loadUser = await UserService.authenticate({
+                address: secondAddress,
+                trust: {
+                    phone,
+                },
+                overwrite: true,
+            });
+
+            const findUser = await sequelize.models.UserModel.findOne({
```
obernardovieira, line 235: you can do
```suggestion
            const findUser = await models.user.findOne({
```
it's easier to find issues if we change either the name or any parameter.

obernardovieira (changes requested):
Some small notes.
Once finished re-request review. Should be good by then.

obernardovieira (approved):

PR #90, opened 2021-08-13
[IPCT1-400] - test inactive accounts on cron job - detect suspicious

dev-jotape:
This PR fixes [IPCT1-400] at https://impactmarket.atlassian.net/browse/IPCT1-400

## Changes
Changed cron job verifyUserSuspectActivity to ignore inactive accounts in suspicious account verification

## Tests
refactoring of verifyUserSuspectActivity.test


dev-jotape:
@obernardovieira I made the change in the verifyUserSuspectActivity service, making a query with the WITH clause. I didn't do many queries like that, so if you have any suggestions to improve it, I'm open.
after WITH clause, it is not possible to run 2 queries, so to get around this I put an update inside the WITH clause and another one after.

obernardovieira:
The `in`, `not in` here has probably a few elements, still, it would be nice if we can avoid it. And also, There's no need in setting someone as a suspect if already suspect.

I didn't test but tried to change the query a little bit
```
WITH
    active_users
        AS (SELECT "user".address, "trust".phone
                FROM "user"
                LEFT OUTER JOIN app_user_through_trust AS "through_trust"
                    ON "user".address = "through_trust"."userAddress"
                LEFT OUTER JOIN app_user_trust AS "trust"
                    ON "through_trust"."appUserTrustId" = "trust".id),
    suspect_phone
        AS (select phone from active_users group by phone HAVING count(phone) > 1)
    UPDATE "user" SET suspect = true WHERE address = active_users.address and suspect_phone.phone = active_users.phone and suspect = false
```

does this make sense?

regarding not updating two in a row, well, I guess there are three options. One is yours, another is two separate queries, and the third is a bigger query with conditions like

```
WITH
    active_users
        AS (SELECT "user".address, "trust".phone
                FROM "user"
                LEFT OUTER JOIN app_user_through_trust AS "through_trust"
                    ON "user".address = "through_trust"."userAddress"
                LEFT OUTER JOIN app_user_trust AS "trust"
                    ON "through_trust"."appUserTrustId" = "trust".id),
    suspect_phone
        AS (select phone, count(phone) n from active_users group by phone)
    UPDATE "user" SET suspect = case when suspect_phone.n > 1 then true else false end
        WHERE address = active_users.address and suspect_phone.phone = active_users.phone
          and suspect = case when suspect_phone.n > 1 then false else true end;
```
(didn't test either)

I wonder what would play best with 1M users (all active) with 1k updates (either turning suspect or stop being suspect). Since they are quite similar, between running two queries and just one but more complex, I think the result would likely be the same. Probably doing the update within the `AS ()` as you've done, might be faster. But that might be confusing later on.

What's your opinion?

PR #98, opened 2021-08-30
[IPCT1-363] - user should be able to delete account

dev-jotape:
This PR fixes [IPCT1-363] at https://impactmarket.atlassian.net/browse/IPCT1-363

## Changes
- unlink tables: managers, beneficiary, and beneficiary transactions from the user table.
- changed services of search and list beneficiaries/managers

## New
Endpoint and cron job to delete users.

## Tests
new test file called verifyDeletedAccounts.
tests to validate the exclusion flow and  after some exclusion, validate the services that use the user info


src/database/migrations/z1629758770-update-stories-trigger.js
```diff
@@ -0,0 +1,41 @@
...
+          UPDATE story_engagement SET loves = loves + 1 WHERE "contentId"=NEW."contentId";
+          RETURN NEW;
+      ELSEIF (TG_OP = 'DELETE') THEN -- DELETE operations
+          UPDATE story_engagement SET loves = loves - 1 WHERE "contentId"=OLD."contentId";
+          RETURN OLD;
+      END IF;
+    END;
+    $$ LANGUAGE plpgsql;
+
+    DROP TRIGGER IF EXISTS update_loves_stories
```
obernardovieira, line 23: No need to drop it. See above, it says `CREATE OR REPLACE FUNCTION`.
You can remove this drop and the create trigger. Unless you are changing when the trigger is triggered (`BEFORE INSERT OR DELETE`) or the table, which I think is not the case.

src/services/app/user.ts
```diff
@@ -23,6 +23,8 @@ export default class UserService {
     public static appUserThroughTrust = models.appUserThroughTrust;
     public static appMediaContent = models.appMediaContent;
     public static appMediaThumbnail = models.appMediaThumbnail;
+    public static storyContent = models.storyContent;
```
obernardovieira: TBH, I've been trying to remove this and instead use `models.storyContent` (same for other models) within each method. Any opinion regarding this? My reasoning was that, at some point, these instances were not used anywhere on the code. Any opinion?

src/worker/jobs/cron/user.ts
```diff
@@ -52,4 +53,84 @@ export async function verifyUserSuspectActivity(): Promise<void> {
...
+
+        await models.manager.update({
+            active: false
+        }, {
+            where: {
+                address: { [Op.in]: addresses }
+            }
+        });
+
+        await models.storyContent.destroy({
```
obernardovieira, line 77: What if, instead of this, we use `ON DELETE: CASCADE`, and so, we only have to delete the user? Same for `storyUserEngagement`. Even for `trust`.

src/database/models/ubi/beneficiary.ts
```diff
@@ -60,11 +60,11 @@ export function initializeBeneficiary(sequelize: Sequelize): void {
...
-                references: {
-                    model: 'user',
-                    key: 'address',
-                },
-                onDelete: 'RESTRICT', // delete only if active = false, separately
+                // references: {
+                //     model: 'user',
+                //     key: 'address',
+                // },
+                // onDelete: 'RESTRICT', // delete only if active = false, separately
```
obernardovieira, line 68: Can we still use `include: [ model: beneficiary ]` from `user` or the opposite, without this? And if so, wouldn't it be confusing? :thinking:

src/services/app/user.ts
obernardovieira: Ahh, seems to be outdated. Sorry :facepalm:

src/services/app/user.ts
```diff
@@ -494,4 +494,51 @@ export default class UserService {
...
+                where: { active: true, address },
+            });
+
+            if (manager) {
+                const managersByCommunity = await this.manager.findAll({
+                    where: {
+                        active: true,
+                        communityId: manager.communityId,
+                    },
+                    include: [
```
obernardovieira, line 510: Should we also consider `where: { deletedAt: null }` ?

src/services/ubi/community.ts
```diff
@@ -628,7 +628,13 @@ export default class CommunityService {
                 active: true,
             },
         });
-        return result.map((r) => r.toJSON() as ManagerAttributes);
+        return result.map((r) => {
+            const manager = r.toJSON() as ManagerAttributes;
+            return {
```
obernardovieira, line 633: By changing this, we have to keep in mind that the frontend will need to know this information. Up until now, the returned information was a `ManagerAttributes` object (see https://github.com/impactMarket/api/blob/master/src/api/routes/community.ts#L659) for reference. As the return structure is changed, the swagger docs need to be updated, so the frontend knows what to update.

src/worker/jobs/cron/user.ts
obernardovieira: Maybe it's possible to call `models.user.destroy` with a given condition, and the database handles the remaining?

src/database/migrations/z1629758770-update-stories-trigger.js
dev-jotape: you're right, I made the change

src/worker/jobs/cron/user.ts
dev-jotape: I made some changes in migrations: storyContent and storyUserEngagement, including ON DELETE: CASCADE.
in this way, I could use only one "destroy". (on user)
but, about UserTrust table, I couldn't use the cascade. So I kept a destroy for the UserTrust table and then I do the destroy on the User table.
cascade worked fine on the app_user_through_trust table

src/database/models/ubi/beneficiary.ts
dev-jotape: yes, we can use include because I kept the associations.
and to delete a user profile and keep the beneficiary/manager (even if the address doesn't refer to any user), I needed to remove this restriction

src/services/app/user.ts
dev-jotape: yes, I included this

src/services/ubi/community.ts
dev-jotape: sure! I added this new information at the swagger, on the Beneficiary and Manager list.

src/worker/jobs/cron/user.ts
obernardovieira: ahh I see. I did some search, but indeed nobody does that. Well, fine, it's done within a transaction. I think there's no issue.

src/database/models/ubi/beneficiary.ts
obernardovieira: Ok. Then, just remove the commented code and write a note (in comment) like `// this is associated with user table` or something like that.

src/services/app/user.ts
obernardovieira: Can you add just one more test to test this case, please?

src/database/models/ubi/beneficiary.ts
dev-jotape: done

src/services/app/user.ts
dev-jotape: Done.
I forgot to add in the authentication endpoint the option to "recover" the account, so I made this change and its respective tests

tests/integration/services/user.test.ts
```diff
@@ -431,6 +442,109 @@ describe('user service', () => {
...
+
+        it('should not delete a user when he is one of the only two managers in the community', async () => {
+            UserService.delete(users[0].address)
+                .catch((err) => expect(err).to.be.equal('Not enough managers'))
+                .then(() => {
+                    throw new Error('expected to fail');
+                });
+        });
+
+        it('manager should be target to be delete', async () => {
```
obernardovieira, line 485:
```suggestion
        it('manager should be able to be delete account', async () => {
```

tests/integration/services/user.test.ts
```diff
@@ -431,6 +442,109 @@ describe('user service', () => {
...
+            const user = await models.user.findOne({
+                where: {
+                    address: users[0].address,
+                },
+            });
+
+            expect(user?.deletedAt).to.be.not.null;
+        });
+
+        it('should not delete a user when he is one of the only two managers that are not in a deletion process', async () => {
```
obernardovieira, line 499: for simplicity/readability
```suggestion
        it('should not delete with only two managers (not in deletion process)', async () => {
```

tests/integration/services/user.test.ts
```diff
@@ -431,6 +442,109 @@ describe('user service', () => {
...
+            UserService.delete(users[2].address)
+                .catch((err) => expect(err).to.equal('Not enough managers'))
+                .then(() => {
+                    throw new Error('expected to fail');
+                });
+        });
+
+        it('beneficiary should be target to be delete', async () => {
```
obernardovieira, line 507:
```suggestion
        it('beneficiary should be able to delete account', async () => {
```

obernardovieira (approved):
Nice. I'm approving with a few suggestions.
Feel free to accept and merge the PR.
Maybe another suggestion is to group the `delete` tests by persona. Like "manager", "beneficiary", "user". Can be done in a separate PR (this one is already long enough). If you think it's good, lets setup a task to work on that.

PR #100, opened 2021-09-03
[IPCT1-365] - user should be able to optionally add an email in profile and subscribe/unsubscribe newsletter

dev-jotape:
This PR fixes [IPCT1-365] at https://impactmarket.atlassian.net/browse/IPCT1-365

## Changes
update "edit user" endpoint to receive an optional email;

## New
- added email on user model
- new endpoint to verify if the user is subscribed to the newsletter
- new endpoint to create or remove a user on hubspot

## Tests
tests added on user.test.ts to validate the newsletter flow.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
         "@interfaces": "dist/interfaces"
     },
     "dependencies": {
+        "@hubspot/api-client": "^4.1.0",
         "@sendgrid/mail": "7.4.4",
         "@sentry/node": "6.7.1",
         "@sentry/tracing": "6.7.1",
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
         "@interfaces": "dist/interfaces"
     },
     "dependencies": {
-        "@hubspot/api-client": "^4.1.0",
+        "@hubspot/api-client": "4.1.0",
         "@sendgrid/mail": "7.4.4",
         "@sentry/node": "6.7.1",
         "@sentry/tracing": "6.7.1",
--- a/src/config/validatenv.ts
+++ b/src/config/validatenv.ts
@@ -75,6 +75,7 @@ function validateEnv() {
         HAS_REDIS_TLS: bool({ default: true }),
         INTERNAL_NOTIFICATIONS: bool({ default: false }),
         MAX_REQUEST_PER_USER: num({ default: 900 }),
+        HUBSPOT_KEY: str(),
     });
 }
 
--- a/src/config/validatenv.ts
+++ b/src/config/validatenv.ts
@@ -75,7 +75,7 @@ function validateEnv() {
         HAS_REDIS_TLS: bool({ default: true }),
         INTERNAL_NOTIFICATIONS: bool({ default: false }),
         MAX_REQUEST_PER_USER: num({ default: 900 }),
-        HUBSPOT_KEY: str(),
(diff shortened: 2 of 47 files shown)


src/services/app/user.ts
```diff
@@ -26,6 +28,8 @@ export default class UserService {
 
     private static profileContentStorage = new ProfileContentStorage();
 
+    public static hubspotClient = new Client({ apiKey: config.hubspotKey });
```
obernardovieira, line 31: maybe? :thinking: 
```suggestion
    private static hubspotClient = new Client({ apiKey: config.hubspotKey });
```

package.json
```diff
@@ -44,6 +44,7 @@
         "@interfaces": "dist/interfaces"
     },
     "dependencies": {
+        "@hubspot/api-client": "^4.1.0",
```
obernardovieira, line 47: Lets always use fixed versions. Gives us more control.
```suggestion
        "@hubspot/api-client": "4.1.0",
```

src/config/validatenv.ts
```diff
@@ -75,6 +75,7 @@ function validateEnv() {
         HAS_REDIS_TLS: bool({ default: true }),
         INTERNAL_NOTIFICATIONS: bool({ default: false }),
         MAX_REQUEST_PER_USER: num({ default: 900 }),
+        HUBSPOT_KEY: str(),
```
obernardovieira, line 78: Assuming it won't be much used in dev, we could give it a default value in dev, so developers don't have to worry about what to write in .env.
Maybe
```suggestion
        HUBSPOT_KEY: str({ devDefault: 'xyz' }),
```

src/services/app/user.ts
dev-jotape: in this case, I used public because I need to import this "hubspotClient" on user.test

src/services/app/user.ts
```diff
@@ -483,6 +487,7 @@ export default class UserService {
             children?: number;
             avatarMediaId?: number;
             pushNotificationToken?: string;
+            email?: string;
```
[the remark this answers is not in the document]
dev-jotape: I had tried to use `UserCreationAttributes`,  but there is a conflict of types.
`Argument of type 'UserCreationAttributes' is not assignable to parameter of type 'Partial<User>'.
  Types of property 'trust' are incompatible.
    Type 'AppUserTrustCreation | undefined' is not assignable to type 'AppUserTrust[] | undefined'.
      Type 'AppUserTrustCreation' is missing the following properties from type 'AppUserTrust[]': length, pop, push, concat, and 26 more.`

I could overcome this error of many ways: passing separated to the update function: `this.user.update({ email, active, pushNotification })` ignoring the `trust` attribute, or even forcing the type `any`: `user.update(user as any)`. But I don't know if this ways to fix is better than receive this fields individually in the function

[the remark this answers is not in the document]
obernardovieira: Well, I'm against `any` (we wouldn't use typescript) and usually those errors happen when I'm missing something. So I decided to try it myself. It can't be `UserCreationAttributes`, because it is an `update` operation. You need to use `User` instead :wink:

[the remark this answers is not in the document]
dev-jotape: done!

obernardovieira (approved):

PR #102, opened 2021-09-10
[IPCT1-137] - create notifications infrastructure

dev-jotape:
This PR fixes [IPCT1-137] at https://impactmarket.atlassian.net/browse/IPCT1-137

## New
Created a notification model.
Created 3 news endpoints.
- list all notifications from a user
- count the unread notifications
- mark all notifications as read

## Tests
news tests added on user.test.ts

### DIFF
--- a/src/database/migrations/z1631138691-create-notification.js
+++ b/src/database/migrations/z1631138691-create-notification.js
@@ -23,7 +23,7 @@ module.exports = {
                 allowNull: true,
             },
             params: {
-                type: Sequelize.JSON,
+                type: Sequelize.STRING(32),
                 allowNull: true,
             },
             read: {
@@ -33,10 +33,6 @@ module.exports = {
             createdAt: {
                 type: Sequelize.DATE,
                 allowNull: false,
-            },
-            updatedAt: {
-                allowNull: false,
-                type: Sequelize.DATE,
             },
         });
     },
--- a/src/database/migrations/z1631138691-create-notification.js
+++ b/src/database/migrations/z1631138691-create-notification.js
@@ -1,0 +1,46 @@
+'use strict';
+
+// eslint-disable-next-line no-undef
+module.exports = {
+    async up(queryInterface, Sequelize) {
+        return queryInterface.createTable('app_notification', {
+            id: {
+                type: Sequelize.INTEGER,
+                autoIncrement: true,
+                primaryKey: true,
+            },
+            address: {
+                type: Sequelize.STRING(44),
+                references: {
+                    model: 'user',
+                    key: 'address',
+                },
+                onDelete: 'CASCADE',
+                allowNull: false,
+            },
+            type: {
+                type: Sequelize.INTEGER,
+                allowNull: true,
+            },
+            params: {
+                type: Sequelize.JSON,
+                allowNull: true,
+            },
+            read: {
+                type: Sequelize.BOOLEAN,
+                defaultValue: false,
+            },
+            createdAt: {
+                type: Sequelize.DATE,
+                allowNull: false,
+            },
+            updatedAt: {
+                allowNull: false,
+                type: Sequelize.DATE,
+            },
+        });
+    },
+    down: (queryInterface) => {
+        return queryInterface.dropTable('app_notification');
+    },
+};
--- a/tests/integration/services/user.test.ts
+++ b/tests/integration/services/user.test.ts
@@ -432,5 +432,58 @@ describe('user service', () => {
 
             expect(userUpdated).to.include(data);
         })
+    });
+
+    describe('notifications', () => {
+        let users: User[];
+
+        before(async () => {
+            users = await UserFactory({ n: 2 });
+            const yesterday = new Date();
+            yesterday.setDate(yesterday.getDate() - 1);
+            await models.appNotification.bulkCreate([
+                {
+                    address: users[0].address,
+                    type: 1,
+                    params: {
+                        communityId: 1
+                    },
+                }, {
+                    address: users[0].address,
+                    type: 2,
+                    params: {
+                        communityId: 1
+                    },
+                }, {
+                    address: users[1].address,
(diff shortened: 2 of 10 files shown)


src/database/migrations/z1631138691-create-notification.js
```diff
@@ -0,0 +1,46 @@
...
+            id: {
+                type: Sequelize.INTEGER,
+                autoIncrement: true,
+                primaryKey: true,
+            },
+            address: {
+                type: Sequelize.STRING(44),
+                references: {
+                    model: 'user',
+                    key: 'address',
```
obernardovieira, line 16: TBH, I've been thinking that anything that is not related to on-chain activity, would be linked to the user id, instead of the address. The address is a 44 chars string. The user id is an integer index. I see many advantages. What's your opinion?

src/database/migrations/z1631138691-create-notification.js
```diff
@@ -0,0 +1,46 @@
...
+                },
+                onDelete: 'CASCADE',
+                allowNull: false,
+            },
+            type: {
+                type: Sequelize.INTEGER,
+                allowNull: true,
+            },
+            params: {
+                type: Sequelize.JSON,
```
obernardovieira, line 26: Any reason to JSON instead of string?
I see the possibilities, but I'm wondering if it's worth the extra work :thinking:

src/database/migrations/z1631138691-create-notification.js
```diff
@@ -0,0 +1,46 @@
...
+            },
+            read: {
+                type: Sequelize.BOOLEAN,
+                defaultValue: false,
+            },
+            createdAt: {
+                type: Sequelize.DATE,
+                allowNull: false,
+            },
+            updatedAt: {
```
obernardovieira, line 37: hmm, the only update is to move it to "read: true". I think it's not worth having an "updatedAt" in this table.

tests/integration/services/user.test.ts
```diff
@@ -442,6 +442,59 @@ describe('user service', () => {
...
+        before(async () => {
+            users = await UserFactory({ n: 2 });
+            const yesterday = new Date();
+            yesterday.setDate(yesterday.getDate() - 1);
+            await models.appNotification.bulkCreate([
+                {
+                    address: users[0].address,
+                    type: 1,
+                    params: {
+                        communityId: 1
```
obernardovieira, line 460: I think I see where your idea for the JSON comes from. So you were thinking about objects rather than a string, so it could host more than one param. I honestly think, there will be only one param.

tests/integration/services/user.test.ts
```diff
@@ -442,6 +442,59 @@ describe('user service', () => {
...
+        });
+
+        it('mark all notifications as read', async () => {
+            await UserService.readNotifications(users[0].address);
+
+            const firstUser = await UserService.getUnreadNotifications(users[0].address);
+            const secondUser = await UserService.getUnreadNotifications(users[1].address);
+
+            expect(firstUser).to.be.equal(0);
+            expect(secondUser).to.be.equal(1);
```
obernardovieira, line 497: I see these tests are dependant on each other. For example, if they are reordered, one of them will fail. It's good practice not to do it.

obernardovieira (changes requested):

src/database/migrations/z1631138691-create-notification.js
dev-jotape: I thought to use JSON to be easier to pass as a parameter in some action on the application. If we saved more than one param, on the app, will be necessary some treatment, maybe making a `split` to extract from a string the params. As an object, we can use it directly.

tests/integration/services/user.test.ts
dev-jotape: I see, in this case, I can change to string

src/database/migrations/z1631138691-create-notification.js
dev-jotape: changed to string!

src/database/migrations/z1631138691-create-notification.js
dev-jotape: updatedAt removed

tests/integration/services/user.test.ts
dev-jotape: changed to string

tests/integration/services/user.test.ts
dev-jotape: I don't know if a understood well. No matter the order, always the users[0] will have 2 notifications (that, in the second test, will be read), and the users[1] always have one notification (that wouldn't be read), because in the `before` I created the notifications based on the user's address.

src/database/migrations/z1631138691-create-notification.js
dev-jotape: it is a good point. 
I was searching about it and really is better (more efficient and simpler) to do comparisons and joins by an integer (ID) than a string.
so, we can have faster queries, especially the queries with the `includes`

tests/integration/services/user.test.ts
obernardovieira: If you run 'mark all notifications as read' before 'get all notifications from a user', it will fail in 'get all notifications from a user' because 'mark all notifications as read' changes all the `users[0]` notifications to read and will fail at line 486, because `read` will be true. Or does `before` run each time before an `it` test. I don't think so. I think it's `beforeEach`. But you don't need to do that. Instead, you can create 3 users and make tests independent.

dev-jotape: got it!
you are right, the `before` runs once before the tests.
I followed your suggestion and used 3 users

obernardovieira:
I think you just forgot to replace the user address with the user id.
I was thinking how would that work (because we don't use it anywhere) and realize it's necessary to return the user id when /auth and /welcome.

Everything else looks good.

PR #103, opened 2021-09-13
[IPCT1-459] - refactor error return

dev-jotape:
This PR fixes [IPCT1-459] at https://impactmarket.atlassian.net/browse/IPCT1-459

## Changes
instead of an error returning only a string, it was changed to an object (Error class) that returns the name and message.


obernardovieira (approved):
This is fantastic. Finally the error will not be of type `any` anymore on the mobile app :relaxed:

PR #106, opened 2021-09-13
[IPCT1-465] - update inflow table to support different assets

dev-jotape:
This PR fixes [IPCT1-465] at https://impactmarket.atlassian.net/browse/IPCT1-465

## Changes
Changed the inflow model
- removed communityId
- included contractAddress, asset and value.
- inflow trigger

Changed the services that use the inflow information, like chainSubscribers, globalDailyState, and cron jobs.

## Tests
All the tests are running with no errors, especially the tests that calculate the metrics.

### DIFF
--- a/src/database/migrations/z1631541043-update-inflow.js
+++ b/src/database/migrations/z1631541043-update-inflow.js
@@ -1,0 +1,48 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    if (process.env.NODE_ENV === 'test') {
+      return;
+    }
+
+    //add column
+    await queryInterface.addColumn('inflow', 'contractAddress', {
+        type: Sequelize.STRING(44),
+        allowNull: false,
+        defaultValue: '0',
+    });
+
+    await queryInterface.addColumn('inflow', 'asset', {
+      type: Sequelize.STRING(4),
+      defaultValue: 'cUSD',
+      allowNull: false,
+    });
+
+    await queryInterface.addColumn('inflow', 'value', {
+      type: Sequelize.DECIMAL(29), // max 9,999,999,999 - plus 18 decimals
+      allowNull: false,
+      defaultValue: 0,
+    });
+
+    // populate new column
+    const query = `
+      UPDATE inflow
+      SET "contractAddress" = community."contractAddress", "value" = "amount"
+      FROM (SELECT "contractAddress", "publicId" FROM community) community
+      WHERE community."publicId" = "communityId"`;
+
+    await queryInterface.sequelize.query(query, {
+      raw: true,
+      type: Sequelize.QueryTypes.UPDATE,
+    });
+
+    // remove old column
+    await queryInterface.removeColumn(
+      'inflow',
+      'communityId'
+    );
+  },
+
+  down: async (queryInterface, Sequelize) => {}
+};
--- a/src/database/migrations/z1631541043-update-inflow.js
+++ b/src/database/migrations/z1631541043-update-inflow.js
@@ -6,7 +6,6 @@ module.exports = {
       return;
     }
 
-    //add column
     await queryInterface.addColumn('inflow', 'contractAddress', {
         type: Sequelize.STRING(44),
         allowNull: false,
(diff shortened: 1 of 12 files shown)


src/database/migrations/z1631541043-update-inflow.js
```diff
@@ -0,0 +1,48 @@
...
+    //add column
+    await queryInterface.addColumn('inflow', 'contractAddress', {
+        type: Sequelize.STRING(44),
+        allowNull: false,
+        defaultValue: '0',
+    });
+
+    await queryInterface.addColumn('inflow', 'asset', {
+      type: Sequelize.STRING(4),
+      defaultValue: 'cUSD',
```
obernardovieira, line 18: I wouldn't do it this way. I think integers here are better, and then we do an enum on code.
Like
`enum Asset {cUSD, cEUR, CELO};`
Reason: I don't see advantages of saving a string on db over using enum on code. But I see advantages on the opposite. It's less likely a bug happens due to some naming issue and it has better performance.
What do you think?

src/database/migrations/z1631541043-update-inflow.js
```diff
@@ -0,0 +1,48 @@
...
+    await queryInterface.sequelize.query(query, {
+      raw: true,
+      type: Sequelize.QueryTypes.UPDATE,
+    });
+
+    // remove old column
+    await queryInterface.removeColumn(
+      'inflow',
+      'communityId'
+    );
```
obernardovieira, line 44: I'm not sure if you need to remove a constraint first. Just writing here, so I don't forget.

src/worker/jobs/cron/community.ts
```diff
@@ -683,9 +683,9 @@ export async function verifyCommunityFunds(): Promise<void> {
                 const community = await CommunityService.getCommunityOnlyById(
                     communityState.communityId
                 );
-                if (community !== null) {
+                if (community !== null && community.contractAddress) {
```
obernardovieira, line 686: Hmm, this `verifyCommunityFunds` cron job has been turned off a while ago because it was sending too many notifications. It made sense initially. We will likely rewrite from scratch. So if you want to delete it rather than update it, please, do so.

src/database/migrations/z1631541043-update-inflow.js
obernardovieira: Also, quick thought (also to not forget). I see you didn't change yet. This change requires updating the triggers.

src/database/migrations/z1631668396-update-triggers.js
```diff
@@ -0,0 +1,36 @@
...
+        await queryInterface.sequelize.query(`
+        CREATE OR REPLACE FUNCTION update_inflow_community_states()
+    RETURNS TRIGGER AS $$
+    declare
+        state_raised numeric(29);
+        -- state_daily_raised numeric(29);
+        n_backer bigint;
+        community_id integer;
+    BEGIN
+        SELECT id INTO community_id FROM community where "contractAddress"=NEW."contractAddress";
```
obernardovieira, line 19: Just thought about one thing. If this is a donation to the DAO, the address does not exist as a community. Is it null then? If so, we can avoid further updates. Also, I just opened a task for us to explore removing this triggers.

src/database/migrations/z1631541043-update-inflow.js
dev-jotape: I didn't have this problem running locally, but, just to make sure, I added the "removeConstraint" before removing the column.
And now, I updated the triggers

src/database/migrations/z1631541043-update-inflow.js
dev-jotape: sure, is a good ideia. I changed to integer

src/worker/jobs/cron/community.ts
dev-jotape: ok, deleted

src/database/migrations/z1631668396-update-triggers.js
dev-jotape: I see. 
I added an IF on the trigger to verify the community_id, if is null, just return and don't do the further updates

obernardovieira (approved):

PR #109, opened 2021-09-17
[IPCT1-409] - read beneficiary rules

dev-jotape:
This PR fixes [IPCT1-409] at https://impactmarket.atlassian.net/browse/IPCT1-409

## Changes
- added column `readRules` on the beneficiary table to save if the beneficiary already read the rules
- added `readBeneficiaryRules` on welcome and auth return 

## New
- endpoint /community/beneficiaries/read-rules (PUT) that set `readRules` as true

## Tests
new tests added on beneficiary.test

### DIFF
--- a/src/database/models/ubi/beneficiary.ts
+++ b/src/database/models/ubi/beneficiary.ts
@@ -15,7 +15,7 @@ export interface BeneficiaryAttributes {
     claimed: string;
     lastClaimAt: Date | null;
     penultimateClaimAt: Date | null;
-    readRules: boolean | null;
+    readRules: boolean;
 
     // timestamps
     createdAt: Date;
--- a/src/database/models/ubi/beneficiary.ts
+++ b/src/database/models/ubi/beneficiary.ts
@@ -15,6 +15,7 @@ export interface BeneficiaryAttributes {
     claimed: string;
     lastClaimAt: Date | null;
     penultimateClaimAt: Date | null;
+    readRules: boolean | null;
 
     // timestamps
     createdAt: Date;
@@ -45,6 +46,7 @@ export class Beneficiary extends Model<
     public claimed!: string;
     public lastClaimAt!: Date | null;
     public penultimateClaimAt!: Date | null;
+    public readRules!: boolean;
 
     // timestamps!
     public readonly createdAt!: Date;
@@ -105,6 +107,11 @@ export function initializeBeneficiary(sequelize: Sequelize): void {
             penultimateClaimAt: {
                 type: DataTypes.DATE,
                 allowNull: true,
+            },
+            readRules: {
+                type: DataTypes.BOOLEAN,
+                allowNull: false,
+                defaultValue: false,
             },
             createdAt: {
                 type: DataTypes.DATE,
--- a/src/services/app/user.ts
+++ b/src/services/app/user.ts
@@ -478,7 +478,11 @@ export default class UserService {
             }
             return null;
         };
+
+        let readBeneficiaryRules = false;
+
         if (beneficiary) {
+            readBeneficiaryRules = beneficiary.readRules;
             community = await getCommunity(beneficiary.communityId);
         } else if (manager) {
             community = await getCommunity(manager.communityId);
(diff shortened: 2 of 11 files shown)


src/services/app/user.ts
```diff
@@ -496,6 +500,7 @@ export default class UserService {
         return {
             isBeneficiary: beneficiary !== null,
             isManager: manager !== null || managerInPendingCommunity,
+            readBeneficiaryRules,
```
obernardovieira, line 503: I was looking at this, thinking about the next steps and I just wonder if we could do this differently.
There are some results here to be deprecated (I've just added tasks for that). And we are returning more beneficiary information. I was thinking about changing the result. What do you think of this

```
return {
    beneficiary,
    manager,
    user
}
```
At this point, `beneficiary` would only contain `blocked`, `readRules` and `communityId`. `manager` would contain `communityId`only. And `user` would contain `suspect` only.
This requires some changes here and some changes in `beneficiary` and `manager` table (to which I added tasks on JIRA) but still, looks more organized and simpler to maintain. Any thoughts?
In my opinion, we could start with this approach right away (now) and one UI side we will only deal with beneficiary.readRules for now. Then we add another task and later we do this change.

obernardovieira (commented):
Hmm, I'm not convinced about the endpoint. `/community/beneficiaries/read-rules`. First, it's under `/community`, but the change is related to the user. Second `beneficiaries` sounds like it's about to do a bulk change.
Maybe a different approach is to use `/user/read-rules/(beneficiary|manager)` (not sure if this is how it's done). What do you think?

src/database/models/ubi/beneficiary.ts
```diff
@@ -15,6 +15,7 @@ export interface BeneficiaryAttributes {
     claimed: string;
     lastClaimAt: Date | null;
     penultimateClaimAt: Date | null;
+    readRules: boolean | null;
```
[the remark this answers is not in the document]
dev-jotape: right. I removed the `| null`

src/services/app/user.ts
dev-jotape: I think that is a good structure to return, more clear. 
I added a comment in all the other attributes to be deprecated

dev-jotape:
about the endpoint, I started creating in "users", but after I changed to "community" because all the other endpoints about beneficiary/manager were there.
but thinking again, I believe it makes sense to be in "users"

src/services/app/user.ts
```diff
@@ -493,18 +497,45 @@ export default class UserService {
...
+        }
+
+        if (beneficiary) {
+            response = {
+                ...response,
+                beneficiary: {
+                    blocked: beneficiary !== null ? beneficiary.blocked : false,
+                    readRules: readBeneficiaryRules,
+                    communityId: community ? community.id : undefined,
+                },
```
obernardovieira, line 522: nah, you don't need all that. My idea was really to make it simpler. Just return it as it is (same for others). Make sure you query just that on the queries above instead of everything. For managers for example you get only the communityId. This community id is now the publicId, but no worries, there are other tasks to replace this. We will keep using the deprecated value in UI in the meantime.
```suggestion
                beneficiary,
```

src/types/endpoints.ts
```diff
@@ -82,16 +82,33 @@ export interface IManagersDetails {
...
-    blocked: boolean;
+    communityId?: number; // TODO: deprecated
+    blocked?: boolean; // TODO: deprecated
     // to users not yet registered, the values below do not exist
-    verifiedPN: boolean | undefined; // TODO: deprecated in mobile-app@1.1.5
-    suspect: boolean | undefined;
+    verifiedPN?: boolean | undefined; // TODO: deprecated in mobile-app@1.1.5
+    suspect?: boolean | undefined; // TODO: deprecated
+    beneficiary?: IBeneficiary;
+    manager?: IManager;
```
obernardovieira, line 110: if you agree with my other suggestion of returning the object right from the query, it can be null, but not undefined.
```suggestion
    beneficiary: IBeneficiary | null;
    manager: IManager | null;
```

src/services/app/user.ts
obernardovieira: also, the `getCommunity` will later be removed.

obernardovieira (commented):
See my comments. What do you think?

src/types/endpoints.ts
```diff
@@ -82,16 +82,33 @@ export interface IManagersDetails {
     };
 }
 
+interface IManager {
+    communityId: number | undefined;
```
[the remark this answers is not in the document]
dev-jotape: done

obernardovieira (approved):

PR #112, opened 2021-09-23
[IPCT1-432] - return number of beneficiaries added by a manager

dev-jotape:
This PR fixes [IPCT1-432] at https://impactmarket.atlassian.net/browse/IPCT1-432

## Changes
changed service `getMembers` to return the number of beneficiaries added by each manager.

## Tests
new test added on community.test


src/database/models/ubi/manager.ts
```diff
@@ -15,6 +15,11 @@ export interface ManagerAttributes {
 
     user?: User;
     community?: CommunityAttributes;
+    beneficiaryRegistry?: [
+        {
+            address: string;
+        }
+    ];
```
obernardovieira, line 22: I know you are querying only the address on the only use case now, but always use the complete interface (https://github.com/impactMarket/api/blob/master/src/interfaces/ubi/ubiBeneficiaryRegistry.ts#L8) even if you don't query it all.

src/services/ubi/community.ts
```diff
@@ -634,6 +636,15 @@ export default class CommunityService {
...
                 },
+                {
+                    model: this.ubiBeneficiaryRegistry,
+                    as: 'beneficiaryRegistry',
+                    attributes: ['address'],
+                    required: false,
+                    where: {
+                        activity: UbiBeneficiaryRegistryType.add,
+                    },
+                },
```
obernardovieira, line 647: This is one of those situations where I would not use the include. The reason is this https://askinglot.com/can-a-left-join-create-duplicates. In fact, it also happens with the thumbnails (include above), but if you look close, there's a `separate: true`. "If true, runs a separate query to fetch the associated instances, only supported for hasMany associations" from https://sequelize.org/master/class/lib/model.js~Model.html#static-method-findAll (params details). With the "separate" it runs a subquery, which reduces de amount of process needed on the db but also on sequelize to properly generate the json object.

In this case, what we want is actually to count how many beneficiaries a manager has added. So I guess, a separate query `count` instead, is better.

src/services/ubi/community.ts
```diff
@@ -645,6 +656,7 @@ export default class CommunityService {
             return {
                 ...manager,
                 isDeleted: !manager.user,
+                beneficiaryRegistry: manager.beneficiaryRegistry?.length,
```
obernardovieira, line 659: `beneficiaryRegistry` doesn't sound right. It's about, how many beneficiaries were added. Not sure what is best, as I'm not good naming things either. maybe `addedBeneficiaries`

tests/integration/services/community.test.ts
```diff
@@ -1259,4 +1262,66 @@ describe('community service', () => {
...
+
+    describe('get manager', () => {
+        afterEach(async () => {
+            await truncate(sequelize, 'Manager');
+            await truncate(sequelize, 'Beneficiary');
+            await truncate(sequelize, 'Community');
+            await truncate(sequelize);
+        });
+
+        it('should return a list of managers', async () => {
```
obernardovieira, line 1274: We need more tests.
I suggest renaming this one to `should return a list of added beneficiaries by current managers`
And also add
`should return a list of added beneficiaries by previous managers (deleted accounts)`
`should return a list of added beneficiaries by previous managers (existing accounts but not managers)`
`should return a list of added beneficiaries by previous managers (managers in a different community)`

obernardovieira (changes requested):

src/database/models/ubi/manager.ts
dev-jotape: done!

src/services/ubi/community.ts
dev-jotape: done!

src/services/ubi/community.ts
dev-jotape: yes.. when I was doing this I thought about the `include`.
But, by `do a separate query` do you mean to add this `separate: true` or write a new query inside the `map` below to count the beneficiaries added by each manager?

tests/integration/services/community.test.ts
dev-jotape: Ok, I wrote the test about the deleted account. But, the service `getManager` returns just the active managers. Do I need to change the service to return all the managers and the count of beneficiaries added? even the managers that were removed?

src/services/ubi/community.ts
obernardovieira: After reading your comment and thinking, I'd go with a separate query. Instead of one by manager, it can be done with a `select manager_address, count(added_addresses) from ubi_beneficiary_registry where manager_address in (the_list_goes_here) group by manager_address` (pseudo-code-query). What do you think?

tests/integration/services/community.test.ts
obernardovieira: ahhh nice catch :thinking: 
And that requires changes on my last PR on the app too :thinking: 
Quick though, what if we add a parameter to filter only active, or only inactive, or both?
Something like `/community/{id}/managers/{query}` and here, with a `filterByActive`, if true, returns only active, if false, return only inactive, if undefined returns both.

This way, it works for the public community data (all results) and manager management in-app, filtered by active/inactive. I'm trying to replace an old endpoint https://github.com/impactMarket/mobile-app/pull/225/files#diff-dae814b85d64f1d8e5f9f7a60fb16d881f0ef980b46aae13f50b7318f427ca1bL69.

How does that sound?

src/services/ubi/community.ts
dev-jotape: right, but make this query is the same that `separate: true` generate, isn't it?
if I use separate option, sequelize make this query and organize it with the JSON.

tests/integration/services/community.test.ts
dev-jotape: sounds great, I added this filter!

src/services/ubi/community.ts
obernardovieira: yes, but it returns the UbiBeneficiaryRegistry and we don't need that for anything, we just need a number. Imagine a community with 1500 beneficiaries. It will basically fetch 1500 results from UbiBeneficiaryRegistry just to count them.

It might be possible though to make this `count` and `group` within the include, using the `separate: true` and that would simplify the code. I've never done that.

obernardovieira: This https://stackoverflow.com/a/63621045/3348623 looks nice, the problem is that you need to use `raw: true`, otherwise the `attributes:[]` still return the primary key, which is the entry id, if I'm not mistaken.

It's worth look at the underlying queries being made, and write some code documentation since this is getting a big complex.

dev-jotape: I decided to make a separate query, following your first recommendation. I tried to make this `count` with the `include` clause but didn't work well when you have others includes in this same query...

src/services/ubi/community.ts
```diff
@@ -643,14 +653,30 @@ export default class CommunityService {
...
+        const address = result.map((el) => `'${el.address}'`);
+        const beneficiariesAdded:
+            | { count: number; from: string }[]
+            | undefined = await this.sequelize.query(
+            `SELECT count("address"), "from" FROM ubi_beneficiary_registry WHERE "from" IN (${address.toString()}) GROUP BY "from"`,
+            {
+                raw: true,
+                type: QueryTypes.SELECT,
+            }
+        );
```
obernardovieira, line 669: If we can avoid raw queries, lets do it. Keeping the code clean and consistent.
Follow this example https://stackoverflow.com/a/66393583/3348623 something like a .findAll with `raw:true` and `fn count` generates de same query, and it's easier to read.

src/services/ubi/community.ts
```diff
@@ -643,14 +653,30 @@ export default class CommunityService {
...
+
         return result.map((r) => {
             const manager = r.toJSON() as ManagerAttributes;
+            const addedBeneficiaries = beneficiariesAdded.find(
+                (el) => el.from === manager.address
+            );
             return {
                 ...manager,
                 isDeleted: !manager.user,
+                addedBeneficiaries: addedBeneficiaries ? Number(addedBeneficiaries?.count) : 0,
```
obernardovieira, line 679: you can use https://www.postgresqltutorial.com/postgresql-coalesce/ on the query to avoid nulls :wink:

src/services/ubi/community.ts
dev-jotape: done!

src/services/ubi/community.ts
dev-jotape: in this case `coalesce` didn't help because if a manager didn't have registries in `ubi_beneficiary_registry` he won't be returned in the query. But I changed this line, I think it's simpler

obernardovieira (approved):

PR #113, opened 2021-09-23
[IPCT1-410] - managers read the rules

dev-jotape:
This PR fixes [IPCT1-410] at https://impactmarket.atlassian.net/browse/IPCT1-410

## New
endpoint `/user/read-rules/manager`

## Tests
new test file added, manager.test

### DIFF
--- a/src/api/routes/user.ts
+++ b/src/api/routes/user.ts
@@ -466,4 +466,34 @@ export default (app: Router): void => {
         authenticateToken,
         userController.readBeneficiaryRules
     );
+
+    /**
+     * @swagger
+     *
+     * /user/read-rules/manager:
+     *   put:
+     *     tags:
+     *       - "user"
+     *     summary: Read manager rules
+     *     responses:
+     *       "200":
+     *         description: OK
+     *         content:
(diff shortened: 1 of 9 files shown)


src/api/routes/user.ts
```diff
@@ -466,4 +466,34 @@ export default (app: Router): void => {
...
+     *                   type: boolean
+     *     security:
+     *     - api_auth:
+     *       - "write:modify":
+     */
+     route.put(
+        '/read-rules/manager',
+        authenticateToken,
+        userController.readManagerRules
+    );
```
obernardovieira, line 498: What if we do `/read-rules/(beneficiary|manager)` and then merge into one endpoint and one controller method which chooses the service?
reference https://expressjs.com/en/guide/routing.html

dev-jotape: Much better!
I changed. Is this the way you thought?

obernardovieira: Yes, it is :relaxed: looks nice

obernardovieira (approved):

PR #115, opened 2021-09-27
[IPCT1-470] - create app_user table

dev-jotape:
This PR fixes [IPCT1-470] at https://impactmarket.atlassian.net/browse/IPCT1-470

## Changes
created new table `app_user` based on the current user table, but adding the column id
replaced the use of the `user` table  by `app_user`

in the file models/app/appUser.ts, the `address` is still a primary key, because,  in the authentication service if is a new user, a registry is created on tables app_user, app_user_trust, and app_user_through_trust by a single query with an `include`. The table app_user_through_trust is associated with the app_user by the column `address`, in the future, when we start to replace the relations with the app_user by the `id` instead of the `address`, we need to change the primary key to the `id` column.

I tried to find something in the sequelize docs to solve this situation in a better way, but I can't. 
Even using the options `foreignKey` and `otherKey` in the config, was not possible to use the create with associations linking another column that isn't a primary key.
https://github.com/sequelize/sequelize/blob/main/docs/manual/advanced-association-concepts/advanced-many-to-many.md
https://github.com/sequelize/sequelize/blob/main/docs/manual/advanced-association-concepts/creating-with-associations.md


obernardovieira (approved):
:clap: :muscle:

PR #119, opened 2021-09-30
fix app_user migration

dev-jotape:
## Changes
I can't do the `insert` into app_user directly, because the order of the columns is different in each environment (local, staging, production).
So, I changed the migration to get the columns dynamically. First, get the columns of the `user` table, and then, make an insert using this columns.

I tried to do it as simples as possible, but, what do you think @obernardovieira ?

### DIFF
--- a/src/database/migrations/z1603324800-create-appUser.js
+++ b/src/database/migrations/z1603324800-create-appUser.js
@@ -1,19 +1,19 @@
 'use strict';
 module.exports = {
     up: (queryInterface, Sequelize) => {
-        return queryInterface.createTable('user', {
+        return queryInterface.createTable('app_user', {
+            id: {
+                type: Sequelize.INTEGER,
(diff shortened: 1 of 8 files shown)


obernardovieira:
I'll suggest doing something different.
First, let's split the migration into two. One with just the table creation, nothing else (I also think you should rename the migration, using the previous timestamp, so it's one of the first to run. And you can delete the older create). And then a second one with the migration. One simple way to go around that is to, within the migration, create a sequelize model for the old table and one for the new one. Query everything from the first, insert on the second. Sequelize handles everything for you.

As an example of how to create models, look at `z1629716371-populate-ubi-beneficiary-registry.js`. The `queryInterface.sequelize.define` creates the model (connecting to the database table).

dev-jotape:
@obernardovieira , done!

obernardovieira (approved):

PR #138, opened 2021-10-27
[IPCT1-540] add new community contract

dev-jotape:
This PR fixes [IPCT1-540] at https://impactmarket.atlassian.net/browse/IPCT1-540

## Changes
changed the `chainSubscriber` to work with both community contracts.
changed beneficiary service to in the `addBeneficiary` decrease the maxClaim, and in the `removeBeneficiary` increase the maxClaim.

## New
new columns added in `ubi_community_contract` (blocked, decreaseStep) and in `manager` (blocked)


src/worker/jobs/chainSubscribers.ts
```diff
@@ -415,9 +529,17 @@ class ChainSubscribers {
...
+                parsedLog = this.ifaceNewCommunity.parseLog(log);
+            } else {
+                parsedLog = this.ifaceCommunity.parseLog(log);
+            }
             if (parsedLog.name === 'ManagerAdded') {
-                const managerAddress = parsedLog.args[0];
+                const managerAddress =
+                    parsedLog.args.length > 1
+                        ? parsedLog.args[1]
+                        : parsedLog.args[0];
```
obernardovieira, line 542: It's better to revert this. With the new contracts we can achieve the same without this.

src/worker/jobs/chainSubscribers.ts
```diff
@@ -218,10 +235,28 @@ class ChainSubscribers {
     async _processCommunityEvents(
         log: ethers.providers.Log
     ): Promise<ethers.utils.LogDescription | undefined> {
-        const parsedLog = this.ifaceCommunity.parseLog(log);
+        let parsedLog: ethers.utils.LogDescription;
+        if (log.address === config.communityContractAddress) {
+            parsedLog = this.ifaceNewCommunity.parseLog(log);
+        } else {
+            parsedLog = this.ifaceCommunity.parseLog(log);
+        }
```
obernardovieira, line 243: hmm, you can't do it this way because every community has a different address.
For the sake of simplicity, lets assume that, the new contracts do not have the manager address on the event, so past events have the same format. We just need to add new ones, all on the same ABI.

obernardovieira (changes requested):

src/interfaces/ubi/ubiCommunityContract.ts
```diff
@@ -33,6 +33,8 @@ export interface UbiCommunityContract {
     maxClaim: string;
     baseInterval: number;
     incrementInterval: number;
+    blocked?: boolean;
+    decreaseStep?: string;
```
obernardovieira, line 37: I also think this should be blocked: false and decreaseStep: 0 by default, simplifying further implementations no need to check possibly undefined or forcing defined)

src/worker/jobs/chainSubscribers.ts
dev-jotape: I didn't understand, the new contract send 2 parameters, and the manager's address come in position 1. The old contract just send one parameter, with the manager's address in position 0

obernardovieira:
Does this last changes make sense to you?

dev-jotape:
@obernardovieira these changes make sense. But, I'm thinking, instead of duplicate the code inside the `catch` can we just populate the `parsedLog` in the catch and then use ternary conditions:
```javascript
try {
    parsedLog = this.ifaceNewCommunity.parseLog(log);
} catch (_) {
    // compatible with older versions
    parsedLog = this.ifaceCommunity.parseLog(log);
}

if (parsedLog.name === 'BeneficiaryAdded') {
    const beneficiaryAddress = parsedLog.args.length > 1 ? parsedLog.args[1] : parsedLog.args[0];
    const managerAddress = parsedLog.args.length > 1 ? parsedLog.args[0] : txResponse.from;
    ...

```

obernardovieira:
> instead of duplicate the code inside the `catch` can we just populate the `parsedLog` in the catch and then use ternary conditions

oh yes, great idea. Please, be my guest, you can do the changes. I think there's something falling, I'll check again if I broke something.

dev-jotape:
@obernardovieira I did the changes to remove the duplicated code

obernardovieira (approved):
All good.
The deadlocks are breaking more often though.

PR #140, opened 2021-11-04
[IPCT1-542] - refactor community list

dev-jotape:
This PR fixes [IPCT1-542] at https://impactmarket.atlassian.net/browse/IPCT1-542

## Changes
service list community changed to:
1 - enable user in a single endpoint filter the `valids` and `pendings` communities.
2- enable user to pass all the fields that he want, using a syntax close of SQL (`fields=id; publicId; contract.maxClaim; contract.baseInterval; state.*`)

### DIFF
--- a/src/services/ubi/community.ts
+++ b/src/services/ubi/community.ts
@@ -315,10 +315,6 @@ export default class CommunityService {
         fields?: string;
         status?: 'valid' | 'pending';
     }): Promise<{ count: number; rows: CommunityAttributes[] }> {
-        if(!query.status) {
-            query.status = 'valid';
-        }
-
         let extendedWhere: WhereOptions<CommunityAttributes> = {};
         const orderOption: OrderItem[] = [];
 
@@ -447,7 +443,7 @@ export default class CommunityService {
         const communitiesResult = await this.community.findAndCountAll({
             attributes,
             where: {
-                status: query.status,
+                status: query.status ? query.status : 'valid',
                 visibility: 'public',
                 ...extendedWhere,
             },
--- a/src/services/ubi/community.ts
+++ b/src/services/ubi/community.ts
@@ -426,20 +426,20 @@ export default class CommunityService {
         let include: Includeable[];
         let attributes: any;
         const exclude = ['email'];
-        if(query.fields) {
+        if (query.fields) {
             const fields = fetchData(query.fields);
             include = this._generateInclude(fields);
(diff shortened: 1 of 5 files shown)


obernardovieira (commented):
Looks good. Lets just add some tests, specifically for the new parameters.

dev-jotape:
@obernardovieira new tests added

src/services/ubi/community.ts
```diff
@@ -1582,4 +1548,156 @@ export default class CommunityService {
...
+                                  fields.thumbnails.length > 0
+                                      ? fields.thumbnails
+                                      : undefined,
+                              model: this.appMediaThumbnail,
+                              as: 'thumbnails',
+                              separate: true,
+                          },
+                      ]
+                    : [],
+            });
```
obernardovieira, line 1595: Thinking of how the frontend will use it, I don't see any advantage on this. Every parameter is necessary and usually, from the array of thumbnails, only one of them (width some chosen size) is used. So I guess, we can just return all values and not use the `fields.thumbnails` as a condition.
I think we can return only a given size, but lets leave that for later.

dev-jotape: got it!
I did the change to return all fields of thumbnails

src/services/ubi/community.ts
```diff
@@ -1582,4 +1548,152 @@ export default class CommunityService {
...
+                    ? [
+                          {
+                              model: this.appMediaThumbnail,
+                              as: 'thumbnails',
+                              separate: true,
+                          },
+                      ]
+                    : [],
+            });
+        }
```
obernardovieira, line 1592:
```suggestion
        if (fields.cover) {
            extendedInclude.push({
                attributes: ['url'],
                model: this.appMediaContent,
                as: 'cover',
                duplicating: false,
                include: [
                      {
                          attributes: ['url', 'width', 'height', 'pixelRatio'],
                          model: this.appMediaThumbnail,
                          as: 'thumbnails',
                          separate: true,
                      },
                  ],
            });
        }
```
What I mean is this. We always need all these fields. And besides these fields, there are only ids, which are not needed on the UI. So, there are not really optional parameters here.
Does it make sense?

dev-jotape:
@obernardovieira done! but I needed to add the field `id` in the `cover` include because the `separated: true` in the thumbnails include uses this id to run a separated query

obernardovieira (approved):
wow, quick and right on point. Amazing. Thank you

PR #147, opened 2021-11-16
[IPCT1-486] - replace community state

dev-jotape:
This PR fixes [IPCT1-486] at https://impactmarket.atlassian.net/browse/IPCT1-486

## Changes
removed triggers on tables:
- beneficiary
- manager

edited triggers on table:
- ubi_claim
- inflow

changed get community state to get the previous state + today's state

## New

- migration created to calculate the totalClaim, totalRaised and totalBeneficiaries (since the beginning)
- migration created to calculate the totalManagers. As we don't have the sum of the managers day by day, I defined the total of managers only in the last record of the ubi_community_daily_state table


src/services/ubi/community.ts
```diff
@@ -727,12 +729,256 @@ export default class CommunityService {
...
+                        active: false,
+                    },
+                },
+            ],
+            where: {
+                id: communityId,
+            },
+            group: ['Community.id'],
+            raw: true,
+        })) as any;
```
obernardovieira, line 944: Now that you've got to this point, simplify.
No need for groups, do nested includes.

My suggestion, get the community contract address, and then do each one of these queries, individually, like `select sum(amount) from inflow where today & community=x`

src/services/ubi/community.ts
```diff
@@ -1597,10 +1843,51 @@ export default class CommunityService {
...
+                ? fields.state.filter(
+                      (el: string) => !stateExclude.includes(el)
+                  )
+                : { exclude: stateExclude }
+            : [];
+        extendedInclude.push({
+            model: this.ubiCommunityState,
+            attributes: stateAttributes,
+            as: 'state',
+        });
```
obernardovieira, line 1880: I know this is going to be tricky, but we must remove this table. Remove all it's instances.

src/services/ubi/community.ts
```diff
@@ -1690,6 +1962,22 @@ export default class CommunityService {
...
+                    date: yesterdayDate
+                },
+                required: false,
+            },
+            // TODO: deprecated, use dailyState
+            {
+                model: this.ubiCommunityState,
+                attributes: { exclude: ['id', 'communityId'] },
+                as: 'state',
+            },
```
obernardovieira, line 1980: Lets really remove it

src/database/migrations/zz-create-triggers.js
```diff
@@ -54,50 +53,48 @@ EXECUTE PROCEDURE update_beneficiaries_community_states();`);
         SELECT SUM(claimed + NEW.amount) INTO beneficiary_claimed FROM beneficiary WHERE "communityId"=community_public_id AND address=NEW.address;
         UPDATE beneficiary SET claimed = beneficiary_claimed WHERE "communityId"=community_public_id AND address=NEW.address;
         -- update total claimed
-        SELECT SUM(claimed + NEW.amount) INTO state_claimed FROM ubi_community_state WHERE "communityId"=NEW."communityId";
-        UPDATE ubi_community_state SET claimed = state_claimed WHERE "communityId"=NEW."communityId";
+        -- SELECT SUM(claimed + NEW.amount) INTO state_claimed FROM ubi_community_state WHERE "communityId"=NEW."communityId";
+        -- UPDATE ubi_community_state SET claimed = state_claimed WHERE "communityId"=NEW."communityId";
```
obernardovieira, line 57: just remove everything we don't use anymore

src/database/migrations/zz-create-triggers.js
```diff
@@ -54,50 +53,48 @@ EXECUTE PROCEDURE update_beneficiaries_community_states();`);
...
+//         UPDATE ubi_community_state SET managers = managers - 1 WHERE "communityId" = community_id;
+//     END IF;
+//     RETURN NEW;
+// END;
+// $$ LANGUAGE plpgsql;
+// CREATE TRIGGER update_managers_community_state
+// BEFORE INSERT OR UPDATE
+// ON manager
+// FOR EACH ROW
+// EXECUTE PROCEDURE update_managers_community_state();`);
```
obernardovieira, line 91: just remove everything we don't use anymore

obernardovieira (changes requested):

src/services/ubi/community.ts
dev-jotape: but, today the app doesn't use this information? It is safe to remove now?

src/services/ubi/community.ts
dev-jotape: getState refactored

src/services/ubi/community.ts
dev-jotape: removed

src/database/migrations/zz-create-triggers.js
dev-jotape: done

src/services/ubi/community.ts
obernardovieira: oh no, not removing removing. Removing the dependency on this table. We still need the state.

src/services/ubi/community.ts
obernardovieira: It does. But I mean, the table needs to be removed, not the state. We need the state.

dev-jotape: for now, I replaced the fields `claimed`, `raised`, `beneficiaries`, and `managers`, and start to save this information in ubi_community_daily_state (like the global daily state).
but, there are some fields only in the ubi_community_state, like `claims`, `backers`, and `removedBeneficiaries`, that's the reason that I maintained these 2 tables.
I was thinking now.. maybe could be better just call the getState function instead of doing this include in the query. What do you think?

obernardovieira: We can do that. I'm still trying to find out how we use two APIs (once we add TheGraph). Either query from backend and return the complete results or we query on frontend.
But yes, we can use getState here instead.

dev-jotape:
@obernardovieira ready to review

obernardovieira:
Feels like we are getting to the other side of the tunnel :eyes:

src/interfaces/ubi/ubiCommunityState.ts
```diff
@@ -40,18 +40,18 @@
  *            description: Number of unique backers since contract inception
  */
 export interface UbiCommunityState {
```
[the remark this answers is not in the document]
dev-jotape: done!
I kept only the interface

src/database/migrations/z1636472208-update-triggers.js
```diff
@@ -0,0 +1,71 @@
...
+        -- if this address never donated, it's a new backer
+        SELECT count(*) INTO n_backer FROM inflow WHERE "from" = NEW."from" AND "contractAddress"=NEW."contractAddress";
+        IF n_backer = 0 THEN
+            UPDATE ubi_community_state SET backers = backers + 1 WHERE "communityId"=community_id;
+        end if;
+        -- update total raised
+        -- SELECT SUM(raised + NEW.amount) INTO state_raised FROM ubi_community_state WHERE "communityId"=community_id;
+        -- UPDATE ubi_community_state SET raised = state_raised WHERE "communityId"=community_id;
+        -- SELECT SUM(raised + NEW.amount) INTO state_daily_raised FROM ubi_community_daily_state WHERE "communityId"=community_id AND date=DATE(NEW."txAt");
+        -- UPDATE ubi_community_daily_state SET raised = state_daily_raised WHERE "communityId"=community_id AND date=DATE(NEW."txAt");
```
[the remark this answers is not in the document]
dev-jotape: done

obernardovieira (approved):

PR #149, opened 2021-11-22
[IPCT1-526] - add rpc fallback

dev-jotape:
This PR fixes [IPCT1-526] at https://impactmarket.atlassian.net/browse/IPCT1-526

## Changes
changed worker to have a fallback.
after 5 tries to connect with Figment RPC, use a secondary provider.
then after 30 minutes, try to connect with the principal provider again.

### DIFF
--- a/src/config/validatenv.ts
+++ b/src/config/validatenv.ts
@@ -27,6 +27,7 @@ function validateEnv() {
         PORT: port({ devDefault: onlyOnTestEnv(5000) }),
         DATABASE_URL: str(),
         CHAIN_JSON_RPC_URL: str({ devDefault: onlyOnTestEnv('xyz') }),
+        CHAIN_JSON_RPC_URL_FALLBACK: str({ devDefault: onlyOnTestEnv('xyz') }),
         JWT_SECRET: str({ devDefault: onlyOnTestEnv('xyz') }),
         CUSD_CONTRACT_ADDRESS: str({ devDefault: onlyOnTestEnv('xyz') }),
         IMPACT_MARKET_CONTRACT_ADDRESS: str({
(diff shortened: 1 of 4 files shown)


src/worker/jobs.ts
```diff
@@ -153,6 +118,118 @@ export default async (): Promise<void> => {
...
+                            Logger.info(
+                                'Conecting with the principal provider'
+                            );
+                            subscribers.stop();
+                            subscribers = startChainSubscriber();
+                            usingFallbackUrl = false;
+                            failedAnswers = 0;
+                            successfullAnswersAfterCrash = 0;
+                            reconnectChainSubscriber();
+                        });
```
obernardovieira, line 163: if it fails, does it retry again in 30 mins?

src/config/validatenv.ts
```diff
@@ -27,6 +27,7 @@ function validateEnv() {
         PORT: port({ devDefault: onlyOnTestEnv(5000) }),
         DATABASE_URL: str(),
         CHAIN_JSON_RPC_URL: str({ devDefault: onlyOnTestEnv('xyz') }),
+        CHAIN_JSON_RPC_URL_FALLBACK: str({ devDefault: onlyOnTestEnv('xyz') }),
```
obernardovieira, line 30:
```suggestion
        CHAIN_JSON_RPC_URL_FALLBACK: str({ devDefault: onlyOnTestEnv('http://localhost:8545') }),
```

obernardovieira: the line above could be the same :thinking:

src/worker/jobs.ts
dev-jotape: yes, after 30 minutes the flow restarts. Trying to connect to the principal RPC, and in case of continuing error, maintain connected with fallback (30 minutes later retry)

obernardovieira: ahh correct, because it actually calls the method again.

obernardovieira (approved):

PR #153, opened 2021-11-23
[IPCT1-541] - prevent private communities from adding stories

dev-jotape:
This PR fixes [IPCT1-541] at https://impactmarket.atlassian.net/browse/IPCT1-541

## Changes
changed service that adds stories to validate before if the community is private or public.
if is private, return an error.


src/services/story.ts
```diff
@@ -62,6 +62,20 @@ export default class StoryService {
             };
         }
         if (story.communityId !== undefined) {
+            const community = await this.community.findOne({
+                where: {
```
obernardovieira, line 66: No need for the entire table.
```suggestion
            const community = await this.community.findOne({
                attributes: ['id'],
                where: {
```

obernardovieira (approved):

PR #158, opened 2021-12-06
[IPCT1-566] - stories endpoint with high failure rate

dev-jotape:
This PR fixes [IPCT1-566] at https://impactmarket.atlassian.net/browse/IPCT1-566

## Changes
change query generated when a user is logged in


dev-jotape:
@obernardovieira if everything sounds good I can paste this commit into the production branch

obernardovieira (approved):
Yes, please.
But it needs to be reverted later. Otherwise, the `BaseError` is never used as it's intended.

PR #186, opened 2022-01-19
[IPCT1-582] - Calculate ssi

dev-jotape:
This PR fixes [IPCT1-582] at https://impactmarket.atlassian.net/browse/IPCT1-582

## Changes
Change `calcuateCommunitiesMetrics` to use subgraph to calculate ssi


packages/core/src/interfaces/ubi/beneficiary.ts
```diff
@@ -86,3 +86,12 @@ export interface BeneficiaryCreationAttributes {
...
 }
+
+export interface BeneficiarySubgraph {
+    claims: number;
+    community: {
+        id: string;
+    };
+    lastClaimAt: number;
+    preLastClaimAt: number;
+}
```
obernardovieira, line 97: maybe I would keep this on a separate file, where, later, might have all TheGraph entities types.

packages/core/src/subgraph/config.ts
```diff
@@ -0,0 +1,9 @@
+import 'cross-fetch/polyfill';
```
obernardovieira, line 1: what is cross-fetch for?

packages/core/src/subgraph/queries/beneficiary.ts
```diff
@@ -0,0 +1,57 @@
+import { gql } from 'apollo-boost';
+
+import { BeneficiarySubgraph } from '../../interfaces/ubi/beneficiary';
+import { client } from '../config';
+
+export const getBeneficiaries = async (
+    communities: string[]
+): Promise<BeneficiarySubgraph[]> => {
```
obernardovieira, line 8: I'm afraid that this end up consuming all the RAM possible

packages/core/src/subgraph/config.ts
dev-jotape: I was having some problems for run the Apollo Client in node.js (without a browser) and found the answer here: https://stackoverflow.com/questions/50688998/using-apolloclient-with-node-js-fetch-is-not-found-globally-and-no-fetcher-pas

basically, to use the apollo client you need to polyfill or provide a fetch method

obernardovieira: :+1: 
add a comment with that, please

packages/core/src/interfaces/ubi/beneficiary.ts
dev-jotape: right, I change to `src/subgraph/interfaces` to be with everything related to subgraph

packages/core/src/subgraph/queries/beneficiary.ts
dev-jotape: added batchs

obernardovieira (commented):
Last changes make them all run sequential, or did I misunderstand it?
The goal is to process a batch in parallel. Eg: 20 being processed at the same time, once they all finish, then, another 20.
I also think it's better if `getBeneficiaries` queries only one community. Simplifies code readability, the query and remove some extra processing

dev-jotape:
you were right, I changed the code to use more promises and process 20 communities at a time.
could you validate again?

obernardovieira (approved):

PR #188, opened 2022-01-21
[IPCT1-634] - Pending communities with proposals

dev-jotape:
This PR fixes [IPCT1-634] at https://impactmarket.atlassian.net/browse/IPCT1-634

## Changes
Using `the graph` get the proposals and return only communities with no open proposals.

## Tests
new tests added

### DIFF
--- a/packages/core/src/services/ubi/community.ts
+++ b/packages/core/src/services/ubi/community.ts
@@ -388,7 +388,7 @@ export default class CommunityService {
             const communityProposals = await this.getOpenProposals();
             extendedWhere = {
                 ...extendedWhere,
-                id: {
+                requestByAddress: {
                     [Op.notIn]: communityProposals,
                 },
             };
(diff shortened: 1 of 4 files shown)


packages/core/src/services/ubi/community.ts
```diff
@@ -1236,7 +1237,49 @@ export default class CommunityService {
             type: QueryTypes.SELECT,
         });
 
-        const results: ICommunityPendingDetails[] = rawResult.map((c) => ({
+        const callData = rawResult.reduce((acc, el) => {
```
obernardovieira, line 1240: Sorry, this is an old deprecated method.
only /list is used now, with `status=pending` to filter by pending.

Once you try to do this within /list, it will break the page count because it will return the wrong number of results. A workaround is needed

obernardovieira (changes requested):

packages/core/src/services/ubi/community.ts
dev-jotape: could you validate again, please?

obernardovieira (approved):

PR #201, opened 2022-02-15
[BUGFIX] - add missing fields

dev-jotape:
No related tasks.

## Changes
added missing fields in `/community/{id}/managers`


dev-jotape:
@obernardovieira `createdAt` added from the manager's table.
when this change is in production, I will ask @psoaresbj to use the `createdAt` from the manager level instead of the user

obernardovieira (approved):

PR #206, opened 2022-02-22
[IPCT1-515] - remove notifiedBacker

dev-jotape:
This PR fixes [IPCT1-515] at https://impactmarket.atlassian.net/browse/IPCT1-515

## Changes
remove notifiedBacker


packages/core/src/database/migrations/z1645570144-delete-notifiedBacker.js
```diff
@@ -0,0 +1,13 @@
...
+// eslint-disable-next-line no-undef
+module.exports = {
+    async up(queryInterface, Sequelize) {
+        if (process.env.NODE_ENV === 'test') {
+            return;
+        }
+        await queryInterface.dropTable('notifiedbacker');
+    },
+
+    down(queryInterface, Sequelize) {},
```
obernardovieira, line 12: Leave this on the "future migrations" folder (later on we can move to migrations folder). Remember that we have pre-load. If we do both these on the same push, it's going to crash the running instance.

obernardovieira (approved):
Move the file and it's good to go

packages/core/src/database/migrations/z1645570144-delete-notifiedBacker.js
dev-jotape: done!

PR #207, opened 2022-02-27
[IPCT1-509] - Media path

dev-jotape:
This PR fixes [IPCT1-509] at https://impactmarket.atlassian.net/browse/IPCT1-509

## Changes
Change services to use `mediaPath` instead `mediaId`.
When an image is uploaded, just save the image path

## Tests
all tests running


obernardovieira (commented):
for routes and validators, leave only the mediaId, don't add mediaPth nor document it yet.

PR #210, opened 2022-02-28
[IPCT1-681] - Update node version

dev-jotape:
This PR fixes [IPCT1-681] at https://impactmarket.atlassian.net/browse/IPCT1-681

## Changes
update node version


package.json
```diff
@@ -14,7 +14,7 @@
     },
     "homepage": "https://github.com/impactMarket/backend#readme",
     "engines": {
-        "node": "14.15.1"
+        "node": "16.13.0"
```
obernardovieira, line 17:
```suggestion
        "node": ">=16"
```
do the same to others

obernardovieira (commented):
Also add an .nvmrc like https://github.com/impactMarket/subgraph/blob/main/.nvmrc
so we can just do `nvm use` on the repository then

obernardovieira (approved):

PR #216, opened 2022-03-07
[IPCT1-654] - Log of changes

dev-jotape:
This PR fixes [IPCT1-654] at https://impactmarket.atlassian.net/browse/IPCT1-654

## Changes
register a log of changes, when:
- when a community is edited
- when a user edits a profile

## Tests
new tests added in user.test file


packages/core/src/services/app/user.ts
```diff
@@ -554,6 +556,13 @@ export default class UserService {
         if (updated[0] === 0) {
             throw new BaseError('UPDATE_FAILED', 'user was not updated!');
         }
+
+        LogService.saveLog(
+            user.address,
```
obernardovieira, line 561: you could use updated[0].id I think. THis goes together with my comment on "why not accepting just id?"

packages/core/src/services/app/appLog.ts
```diff
@@ -0,0 +1,109 @@
...
+    community: {
+        id: number,
+        name: string,
+    },
+    logs: AppLog[]
+}
+
+export default class LogService {
+    public static async saveLog(
+        user: string | number,
```
obernardovieira, line 16: why not accept just id?

obernardovieira (changes requested):

dev-jotape:
@obernardovieira I made a refactor, could you validate again, please?

obernardovieira (changes requested):

packages/core/src/services/app/user/log.ts
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+
+    public async get(ambassadorAddress: string, type: string, entity: string): Promise<AppLog[]> {
+        const community = await models.community.findOne({
+            attributes: ['id', 'contractAddress'],
+            where: {
+                ambassadorAddress,
+            }
+        });
```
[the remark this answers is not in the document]
dev-jotape: right! I changed a little the function

obernardovieira (approved):

PR #230, opened 2022-03-23
user rules status

obernardovieira:
## Changes
return user rules status

## Tests
no new tests


dev-jotape:
@obernardovieira , approved, just need to fix the conflict.

PR #320, opened 2022-06-20
[IPCT1-743] - use the graph subscriptions to add notification

dev-jotape:
This PR fixes [IPCT1-743] at https://impactmarket.atlassian.net/browse/IPCT1-743

## Changes
use the graph subscription to add notification.

### DIFF
--- a/packages/worker/src/subscription.ts
+++ b/packages/worker/src/subscription.ts
@@ -1,21 +1,26 @@
-import { database, config, interfaces, subgraph, utils } from '@impactmarket/core';
-import ethers from 'ethers';
+import {
+    database,
+    config,
+    interfaces,
+    subgraph,
+    utils,
+} from '@impactmarket/core';
 import { execute } from 'apollo-link';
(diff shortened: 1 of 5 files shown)


packages/worker/src/subscription.ts
```diff
@@ -0,0 +1,310 @@
...
+        const COMMUNITY_QUERY = gql`
+            subscription {
+                communityEntities(
+                    orderBy: startDayId
+                    orderDirection: desc
+                    where: {
+                      startDayId_gte: ${
+                          (new Date().getTime() / 1000 / 86400) | 0
+                      }
+                    }
```
obernardovieira, line 37: won't this make users get notified twice if the server is restarted?

dev-jotape: as there is no DateTime field, only the day, just below there is a check to see if there is already a notification created for the community, just to avoid a situation like this of the server restarting and users being notified of the communities opened today

obernardovieira (approved):

PR #328, opened 2022-06-28
[IPCT1-746] - Stories with multiple images

dev-jotape:
This PR fixes [IPCT1-746] at https://impactmarket.atlassian.net/browse/IPCT1-746

## Changes
- add stories with multiple images
- generate multiples URLs to upload image
- return stories with an array of images

### DIFF
--- a/packages/api/src/routes/v2/story.ts
+++ b/packages/api/src/routes/v2/story.ts
@@ -24,7 +24,7 @@ export default (app: Router): void => {
      *         schema:
      *           type: string
      *         required: true
-     *         description: media mimetype
+     *         description: media mimetype, separated by semicolon (png;jpg).
      *     responses:
      *       "200":
      *         description: OK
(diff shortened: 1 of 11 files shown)


packages/api/src/routes/v2/story.ts
```diff
@@ -24,7 +24,7 @@ export default (app: Router): void => {
      *         schema:
      *           type: string
      *         required: true
-     *         description: media mimetype
+     *         description: media mimetype, separated by semicolon (png;jpg).
```
obernardovieira, line 27: I think it costs the same for frontend (maybe easier) and is easier for backend, if you accept something like `?mime=png&mime=jpg&mime=jpg` because it is transformed in an array when received by the server. Same as we did on story list on v2 where you could filter by country using `?country=PT&country=BR` or by many communities like `?community=2&community=5`. Keeps code cleaner and consistent 💪

dev-jotape: much better in this way!
fixed.

obernardovieira (approved):

PR #353, opened 2022-07-19
351 - refactor pending community

### LINKED ISSUE #351: refactor pending community details
I suggest replacing the UbiCommunityContract with a new table. (cloning existing values, so we don't break the app)
Reason:  We don't use the contract details anymore once the community is accepted. Also we want to save ambassador registry while community is pending (but not after that) only. And allow to change decreaseStep, minTranche, maxTranche as well as currency (not yet available on contracts, but will be soon)

Also, another good point would be to use float values, human readable, and not those big ones with 18 decimals anymore.

Also set the "review" of all valid communities to "accepted" (or whatever is the value, can't remember exactly now), so the existing ones don't show the "claim" option.

dev-jotape:
This PR fixes #351

## Changes
- fix list communities by ambassador
-  add new contract fields (minTranche, maxTranche) and able to edit
- use float values, human-readable, instead of big numbers with 18 decimals. Make sure it doesn't break the app.

## Tests
all tests passed


obernardovieira (approved):
👏 💪

PR #377, opened 2022-08-08
[245] - filter communities by requestedByAddress

dev-jotape:
This PR fixes #245

## Changes
On community requests, filter by the user requesting ("requestedByAddress")

### DIFF
--- a/packages/core/src/services/ubi/community/list.ts
+++ b/packages/core/src/services/ubi/community/list.ts
@@ -23,7 +23,6 @@ export class CommunityListService {
     public async list(query: {
         orderBy?: string;
         filter?: string;
-        name?: string; // TODO: remove
         search?: string;
         country?: string;
         excludeCountry?: string;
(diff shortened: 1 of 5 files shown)


packages/core/src/utils/util.ts
```diff
@@ -275,3 +277,21 @@ export function createThumbnailUrl(
 
     return avatar;
 }
+
+export const getSearchInput = (searchInput: string) => {
+    if (isAddress(searchInput)) {
+        return {
+            address: ethers.utils.getAddress(searchInput),
```
obernardovieira, line 284: try using just
```suggestion
            address: getAddress(searchInput),
```

packages/core/src/services/ubi/community/list.ts
```diff
@@ -80,13 +81,21 @@ export class CommunityListService {
...
-                name: {
-                    [Op.iLike]: `%${query.name}%`,
-                },
-            };
+        if (query.search) {
+            const input = getSearchInput(query.search);
+            if (input.address) {
+                extendedWhere = {
+                    ...extendedWhere,
+                    requestByAddress: query.search,
```
obernardovieira, line 89: shouldn't you use the sanitized here instead? Like `input.search`

obernardovieira (approved):

PR #428, opened 2022-11-14
calculate community metrics

### LINKED ISSUE #423: refactor cronjob for ssi and ubi duration
Should be based on data from TheGraph

SSI will be removed

dev-jotape:
This PR fixes #423 

## Changes
- Refactored and created a new lambda to calculate community metrics and community global metrics.
- Remove the old cron job (calcuateCommunitiesMetrics, calcuateGlobalMetrics)
- Remove all metrics related with SSI (ssi, avgMedianSSI, ssiDayAlone)
- Remove all metrics related with reachOut (reachOut, totalReachOut)
- Remove reachedAddressService, and get data from subgraph and last global daily state
- update webpack config


obernardovieira:
I believe you are still doing the updates.
From my point of view, what we want is to save ubi rate and ubi duration. Everything else can be zero on the database (we will drop it later). Then, build v2 endpoints that merge this ubi rate and ubi duration together with the subgraph data and return just like we do in v1 (can be a slightly different format of course). Nevermind about v1 endpoints. Only the dashboard uses them and they are about to be changed.

PR #473, opened 2022-12-27
node 18

obernardovieira:
It was necessary because of Heroku stack 22, see [here](https://help.heroku.com/DHJMMZ51/fix-node-js-openssl-issues-on-heroku-22). But upon updating, `serverless-bundle` dependency failed because it depends on `webpack-concat-text-plugin`, which requires node < 16 and has been deprecated. [This](https://github.com/AnomalyInnovations/serverless-bundle/pull/343) seems to fix the issue, so, for the time being, the dependency links to that exact PR.


obernardovieira:
Important note, after these changes, building with heroku stack 20 resulted in `segment fault`. It only worked after updating the stack.

PR #522, opened 2023-02-08
cashout provider

### LINKED ISSUE #456: cashout providers registry
setup registry and endpoints to return cashout providers. There are two types of cashout. 1. merchants (which might accept cash in and/or cashout) and exchanges (local or global). It should be possible to filter by country and/or nearby user's location.

See design for reference.
Relates to https://github.com/impactMarket/libera/issues/208

dev-jotape:
This PR fixes #456 

## Changes
Create exchange registry and endpoint to list cashout providers

## Tests
new tests created

### DIFF
--- a/packages/api/src/controllers/v2/generic.ts
+++ b/packages/api/src/controllers/v2/generic.ts
@@ -1,10 +1,15 @@
 import { getAddress } from '@ethersproject/address';
-import { database } from '@impactmarket/core';
+import { database, services } from '@impactmarket/core';
 import { Request, Response } from 'express';
 
 import { standardResponse } from '../../utils/api';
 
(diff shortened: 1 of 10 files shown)


obernardovieira (changes requested):
This is perfect. wallet side we just one more thing. On the `ExchangeRegistry` please add a field name like `customImplementation`, type varchar(16). This is because, we will have some custom integrations for cashout, and would need to develop some things on the wallet side. But when the user open that open we need to now if we show exchange data or the custom cashout implementation.

dev-jotape:
done! @obernardovieira

obernardovieira (approved):
👌

PR #524, opened 2023-02-16
[482] - refactor chainSubscriber

### LINKED ISSUE #482: based on the worker, create listeners, listening only to communityAdmin contract, waiting for CommunityAdded and CommunityRemoved events
Should be incorporated within the API code, running at start, removing the need for the worker

dev-jotape:
This PR fixes #482 

## Changes
Refactor chain subscriber
- move to API
- change JsonRpcProvider by WebSocketProvider


packages/api/src/server.ts
```diff
@@ -136,6 +138,30 @@ export default (app: express.Application): void => {
...
+                    message: 'invalid payloads',
+                    details: error.details.get('query')
+                        ? error.details.get('query').details
+                        : error.details.get('body').details,
+                },
+            });
+        }
+        next();
+    });
+
```
obernardovieira, line 164: I think this was removed before (by me, if I'm not mistaken because I didn't see it being used). What was the reason to add it back?

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,181 @@
...
+                    }
+                );
+                result = parsedLog;
+            }
+        } else if (parsedLog.name === 'CommunityAdded') {
+            const communityAddress = parsedLog.args[0];
+            const managerAddress = parsedLog.args[1];
+
+            const _pCommunity = await database.models.community.findOne({
+                attributes: ['id', 'publicId'],
```
obernardovieira, line 145:
```suggestion
                attributes: ['id'],
```

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,181 @@
...
+        let result: ethers.utils.LogDescription | undefined = undefined;
+
+        if (parsedLog.name === 'CommunityRemoved') {
+            const communityAddress = parsedLog.args[0];
+            const community = await database.models.community.findOne({
+                attributes: ['publicId'],
+                where: { contractAddress: communityAddress },
+            });
+
+            if (!community || !community.publicId) {
```
obernardovieira, line 114:
```suggestion
            if (!community || !community.id) {
```

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,181 @@
...
+    async _processCommunityAdminEvents(
+        log: ethers.providers.Log
+    ): Promise<ethers.utils.LogDescription | undefined> {
+        const parsedLog = this.ifaceCommunityAdmin.parseLog(log);
+        let result: ethers.utils.LogDescription | undefined = undefined;
+
+        if (parsedLog.name === 'CommunityRemoved') {
+            const communityAddress = parsedLog.args[0];
+            const community = await database.models.community.findOne({
+                attributes: ['publicId'],
```
obernardovieira, line 110:
```suggestion
                attributes: ['id'],
```

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,181 @@
...
+                await database.models.beneficiary.update(
+                    {
+                        active: false,
+                    },
+                    {
+                        where: {
+                            communityId: community.publicId,
+                        },
+                    }
+                );
```
obernardovieira, line 137: You can remove this. We don't consume from this table anymore and it will be dropped.

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,181 @@
...
+                where: {
+                    requestByAddress: managerAddress[0],
+                },
+            });
+            for (let index = 0; index < managerAddress.length; index++) {
+                await services.ubi.ManagerService.add(
+                    managerAddress[index],
+                    _pCommunity!.id
+                );
+            }
```
obernardovieira, line 155: You can remove this. We don't consume from this table anymore and it will be dropped.

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    _setupListener(provider: ethers.providers.JsonRpcProvider) {
+        utils.Logger.info('Starting subscribers...');
+        const filter = {
+            topics: this.filterTopics,
+        };
+
+        provider.on(filter, async (log: ethers.providers.Log) => {
+            await this._filterAndProcessEvent(provider, log);
+            services.app.ImMetadataService.setLastBlock(log.blockNumber);
```
obernardovieira, line 86: Lets refactor this to update the last block only each 16560 (23*60*12) blocks, roughly the number of blocks each 23 hours (so it's not always at the same time in the day). We should save this number of course, but maybe save it on redis and each 23 hours save it to the database.

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,181 @@
...
+    async _runRecoveryTxs(
+        startFromBlock: number,
+        provider: ethers.providers.JsonRpcProvider
+    ) {
+        utils.Logger.info('Recovering past events...');
+        const rawLogs = await provider.getLogs({
+            fromBlock: startFromBlock,
+            toBlock: 'latest',
+            topics: this.filterTopics,
+        });
```
obernardovieira, line 57: There's a possible problem here. The provider is a websocket.
I'd suggest a slight change. On the constructor, don't pass any parameters and instantiate the provider within the class. Read from env and create a websocket with websocket url and another provider with http just to use this getLogs. Or maybe, because it's used outside, pass both the http and websocket

packages/api/src/app.ts
```diff
@@ -24,6 +24,9 @@ export async function startServer() {
     await serverLoader(app);
     utils.Logger.info('📡 Express server loaded');
 
+    subscriber.start();
```
obernardovieira, line 27: Perfect, don't await this! 👌

obernardovieira (changes requested):
Minor things and I think we are good to go. Well done 👌

packages/api/src/server.ts
obernardovieira: ahhh I see. Makes sense.
then I think, you can remove the `utils.Logger.error` part, and just add a comment explaining why this is important.

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,170 @@
...
+        utils.Logger.info('Recovering past events...');
+        let startFromBlock: number;
+        let lastBlockCached = await database.redisClient.get('lastBlock');
+        if (!lastBlockCached) {
+            startFromBlock = await services.app.ImMetadataService.getRecoverBlock();
+        } else {
+            startFromBlock = parseInt(lastBlockCached)
+        }
+
+        const rawLogs = await provider.getLogs({
```
obernardovieira, line 58: Maybe I've not made myself clear before. In this provider.getLogs, we should use an http, so the provider should be a JSONRPC and not a websocket. Websockets are for real-time events. Here we are doing a one time request of past events.

packages/core/src/subscriber/chainSubscribers.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    _setupListener(provider: ethers.providers.JsonRpcProvider) {
+        utils.Logger.info('Starting subscribers...');
+        const filter = {
+            topics: this.filterTopics,
+        };
+
+        database.redisClient.set('blockCount', 0);
+
+        provider.on(filter, async (log: ethers.providers.Log) => {
```
obernardovieira, line 91: Ahh, you might have mixed actually. This provider in `provider.on` is the on that should be a websocket.

packages/core/src/subscriber/index.ts
```diff
@@ -0,0 +1,182 @@
...
+let providerFallback: ethers.providers.WebSocketProvider;
+
+export const start = async (): Promise<void> => {
+    provider = new ethers.providers.WebSocketProvider(config.webSocketUrl);
+    providerFallback = new ethers.providers.WebSocketProvider(
+        config.webSocketUrlFallback
+    );
+    subscribers = startChainSubscriber();
+
+    process.on('unhandledRejection', (error: any) => {
```
obernardovieira, line 27: I actually don't know if it still works in case the provider used is a websocket. Did you tried.
I saw this https://github.com/ethers-io/ethers.js/issues/1053
This https://github.com/ethers-io/ethers.js/issues/1053#issuecomment-861059758 seems simple and might do the work.

dev-jotape:
@obernardovieira now Im using a JsonRPCProvider to `getLogs`. Could you validate again please?

obernardovieira (approved):
🥳

PR #534, opened 2023-03-08
update get cashout provider

### LINKED ISSUE #530: refactor get cashout providers

dev-jotape:
This PR fixes #530 

## Changes
Refactor `getCashoutProvider` service


obernardovieira (approved):
I see `allowNull: false` on this new fields, but on testnet I have some. How would it behave? Just testnet, mainnet is empty. Optionally we can remove and re-add, they are only two.

obernardovieira:
You know, actually, one thing that could be done is adding them on a _seeders_ script 🤔

PR #544, opened 2023-03-20
[541] - add control of available rewards

### LINKED ISSUE #541: add reward available per L&E level

dev-jotape:
This PR fixes #541 

## Changes
Add a new field `rewardLimit`, to control how many rewards are available to a level.

### DIFF
--- a/packages/core/src/services/learnAndEarn/claimRewards.ts
+++ b/packages/core/src/services/learnAndEarn/claimRewards.ts
@@ -3,6 +3,7 @@ import { JsonRpcProvider } from '@ethersproject/providers';
 
 import { config } from '../../..';
 import { models, sequelize } from '../../database';
+import { fn, col } from 'sequelize';
 
 const iface = new Interface([
     'event RewardClaimed(address indexed beneficiary, uint256 indexed levelId)',
@@ -28,6 +29,7 @@ export async function registerClaimRewards(
 
         sequelize
             .transaction(async (t) => {
+                let levelId: number;
                 for (let index = 0; index < events.length; index++) {
                     const event = events[index];
 
@@ -38,6 +40,7 @@ export async function registerClaimRewards(
                         continue;
                     }
 
+                    levelId = event.args.levelId;
                     await models.learnAndEarnPayment.update(
                         { status: 'paid' },
                         {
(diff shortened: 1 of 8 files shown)


packages/core/src/services/learnAndEarn/claimRewards.ts
```diff
@@ -60,3 +64,40 @@ export async function registerClaimRewards(
...
+
+    const payments = await models.learnAndEarnPayment.findOne({
+        attributes: [
+            [fn('sum', col('amount')), 'amount'],
+        ],
+        where: {
+            levelId,
+            status: 'paid'
+        }
+    });
```
obernardovieira, line 88: as well, you can use _sum_

packages/core/src/services/learnAndEarn/list.ts
```diff
@@ -233,12 +233,44 @@ export async function listLessons(userId: number, levelId: number) {
...
+
+    const payments = await models.learnAndEarnPayment.findOne({
+        attributes: [
+            [fn('sum', col('amount')), 'amount'],
+        ],
+        where: {
+            levelId,
+            status: 'paid'
+        }
+    });
```
obernardovieira, line 269: as well, _sum_

packages/core/src/services/learnAndEarn/list.ts
```diff
@@ -233,12 +233,44 @@ export async function listLessons(userId: number, levelId: number) {
...
+
+    if (!payments?.amount) {
+        return level.rewardLimit
+    }
+
+    return level.rewardLimit - payments.amount;
```
obernardovieira, line 275: This is tricky to get to exactly zero. And the frontend only needs a boolean. So, I'd suggest, returning true only if the remaining amount is bigger than the max rewards a user can get. To avoid leaving tokens to disburse, what we can do is, when a user finishes a level, we make a calculation to see if the total to receive is less that the amount available, and if so, generate the signature.

This is still not ideal, but we can iterate later. And also, this can be improved with some [subgraph entities](https://github.com/impactMarket/subgraph/issues/81).

obernardovieira (changes requested):

packages/core/src/services/learnAndEarn/claimRewards.ts
dev-jotape: done

obernardovieira (approved):
🤸

Issue #540, opened 2023-03-14
send push notification when user added as borrower


obernardovieira:
To notify when added as borrower, we can use the chain subscribers, as we do now for other things.
Repayment dates for now are always each 30 days since the user has got the funds. We will have that data on a subgraph.

PR #552, opened 2023-03-29
send push notification when user added as borrower

dev-jotape:
This PR fixes [IPCT1-X] at #540 

## Changes
Send push notification when a user added as borrower


packages/core/src/utils/pushNotification.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export async function sendFirebasePushNotification(title: string, body: string, token: string) {
+    try {
+        const message = {
+            notification: {
+                title,
+                body
+            },
+            token,
```
obernardovieira, line 70: You can send bulk notifications.
You can use the example provided on the wallet https://github.com/impactMarket/libera/blob/main/fcm-server-example.js Limit is 500 tokens per request. 1501 would be 3 requests.

packages/core/src/utils/pushNotification.ts
```diff
@@ -0,0 +1,86 @@
...
+        await axios.post(
+            'https://fcm.googleapis.com/fcm/send',
+            JSON.stringify(message),
+            {
+                headers: {
+                    Authorization: 'key=' + config.firebaseKey,
+                    'Content-Type': 'application/json'
+                }
+            }
+        );
```
obernardovieira, line 82: It might be simpler if you use the firebase-admin and this example https://github.com/impactMarket/libera/blob/main/fcm-server-example.js

packages/core/src/utils/pushNotification.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    // filter users that have walletPNT or appPNT
+    users = users.filter(el => el.walletPNT || el.appPNT);
+
+    // get only unique languages
+    const languages = [...new Set(users.map(el => el.language))];
+
+    // get notification from prismic
+    const prismicNotifications = {};
+    const promises = languages.map(async language => {
```
obernardovieira, line 29:
```suggestion
    const fetchMessagesFromPrismic = languages.map(async language => {
```

Suggesting this naming because I first got confused with what was happening here and was about to comment something very non-sense 🤣

packages/core/src/utils/pushNotification.ts
```diff
@@ -0,0 +1,86 @@
...
+    // registry notification
+    await models.appNotification.bulkCreate(users.map(el => ({
+        userId: el.id,
+        type,
+        isWallet: true,
+        isWebApp: true,
+    })));
+
+    // filter users that have walletPNT or appPNT
+    users = users.filter(el => el.walletPNT || el.appPNT);
```
obernardovieira, line 22: Actually, you have to filter by wallet. App push notifications are not setup, and they don't use FCM. I'm actually not sure if it even works.

packages/core/src/utils/pushNotification.ts
```diff
@@ -0,0 +1,86 @@
...
+export async function sendNotification(
+    users: AppUserModel[],
+    type: NotificationType,
+) {
+    // registry notification
+    await models.appNotification.bulkCreate(users.map(el => ({
+        userId: el.id,
+        type,
+        isWallet: true,
+        isWebApp: true,
```
obernardovieira, line 18: So far, all push notifications go to wallet and app, but likely won't be the case in the future. It's better if you provide options on the method.

packages/core/src/utils/pushNotification.ts
```diff
@@ -0,0 +1,86 @@
...
+    // send notification
+    const promises2 = users.map(async user => {
+        let prismicData = prismicNotifications[user.language];
+
+        if (!prismicData) {
+            prismicData = prismicNotifications['en'];
+        }
+
+        sendFirebasePushNotification(prismicData.title, prismicData.description, (user.walletPNT || user.appPNT)!);
+    });
```
obernardovieira, line 58: related to the message below. You can group up to 500 per language and submit them all at once. It will be 1 request instead of 500

packages/core/src/contracts/MicrocreditABI.json
```diff
@@ -0,0 +1,635 @@
...
+      {
+        "indexed": false,
+        "internalType": "uint256",
+        "name": "claimDeadline",
+        "type": "uint256"
+      }
+    ],
+    "name": "LoanAdded",
+    "type": "event"
+  },
```
obernardovieira, line 44: you can remove everything else and leave just this (within [ ])

obernardovieira (changes requested):

dev-jotape:
@obernardovieira , can you review it again, please?

packages/core/src/utils/pushNotification.ts
```diff
@@ -0,0 +1,112 @@
...
+    // send notification
+    const submitPushNotificationsToFCM = users.map(async user => {
+        let prismicData = prismicNotifications[user.language];
+
+        if (!prismicData) {
+            prismicData = prismicNotifications['en'];
+        }
+
+        sendFirebasePushNotification(prismicData.title, prismicData.description, users.map(el => el.walletPNT!));
+    });
```
obernardovieira, line 66: wait, I'm not sure I understood this. Does this iterate over each user and sends push notifications to all users? Considering 50 users, wouldn't this send 50 notifications to each?

dev-jotape: you are right... I refactored this block to send notifications by group of languages

obernardovieira (approved):

PR #575, opened 2023-04-20
add/list admin levels and update webhook

### LINKED ISSUE #565: [LaE] add level admins
A given account should be able to submit requests to new level, generating a new level id entry. Should also be able to see progress (users that took the level, that finished it and the funds)

* create new level
* list levels

dev-jotape:
This PR fixes #565 

## Changes
- create 2 new tables (`learn_and_earn_prismic_level` and `learn_and_earn_prismic_lesson`)
- update `learn_and_earn_level` to set "old" columns as optional and add new columns
- update webhook.
- L&E tests commented because need to be redone in features tasks

The new tables will link the levels and lessons saved on the database with Prismic.


obernardovieira (approved):
Great 🎉 

Some notes. I would probably not control `isLive` on the `learn_and_earn_prismic_lesson` or level. Or at least would not control with prismic. It's great for data, but I don't think we should rely on it to set variables.
On the "create level", you've included a title. I'm not sure it I would call it title. Because the title is in prismic. If you are thinking about the end user, I'd say, we can give it the level id for now. And later, if we feel the need, maybe we add some sort of slug or something, for guidance purpose only. But I wouldn't worry much.
I'd probably call `listLevelsByAdmin` instead of `listAdminLevels`. Sounds like something else.
Finally, I'm just a bit lost on how the `learn_and_earn_prismic_lesson` connects to the user level. If you are thinking of the level_id + lesson_id and the lessonId, you probably need to change `lessonId: prismicLesson.data.id` to `lessonId: `${prismicLevel.data.id}${prismicLesson.data.id}`. Prismic content creators should be writing 1,2,3,4 not 71,72,73,74 (if it was level 7 for example). Or did I miss something?

dev-jotape:
About the `isLive`, for now, I'll keep it this way because it's good to control levels/lessons in production. Later we can just remove this field on Prismic and control only on the database.
`title` removed. Indeep I was thinking about the end user (who submitted the new level).
replaced `listAdminLevels` by `listLevelsByAdmin`.
And finally about the `learn_and_earn_prismic_lesson`, the first idea was to create a column joining the levelId and lessonId (called referenceCode), but in this way, it's challenging to do some queries (creating an association between tables). So, I keep it in 2 separate columns, in order to associate easily `learn_and_earn_prismic_lesson` and `learn_and_earn_level`.

PR #579, opened 2023-05-04
L&E - change lesson interval

dev-jotape:
No task.

## Changes
Update flag `completedToday` returned on list lessons, to be true only if the user already completed a lesson in the same level.
Thus allowing users to do lessons in parallel.


obernardovieira (approved):
oh wow, you are fast 💙 🎉

PR #606, opened 2023-05-17
Add rules on levels

### LINKED ISSUE #563: [LaE] add support to target users
Lets suppose we want to target users just in countries A,B,C. Or we want to target only 5k users. Or just beneficiaries.
A given level should be only displayed if those conditions are met. Only if user is in country A,B,C (based on profile information and the country of the validated phone number).
In the case of limit of users, that limit should be considering by, number of people starting a level in the last X days that haven't finished + number of users that have finished.

This task is dependant on https://github.com/impactMarket/backend/issues/562

dev-jotape:
This PR fixes #563 

## Changes
Added column `rules` on `learn_and_earn_level` to add more control over the courses.
On the level list, these rules are used to filter levels.


packages/core/src/database/models/learnAndEarn/learnAndEarnLevel.ts
```diff
@@ -20,6 +20,11 @@ export class LearnAndEarnLevelModel extends Model<
     public asset!: string;
     public adminUserId!: number;
     public status!: string;
+    public rules!: {
+        countries: string[];
+        roles: string[];
+        limitUsers: number;
+    };
```
obernardovieira, line 27: If it was a simple read field, I'd be fine with this. But we have to apply filters here. I think this can get way too complex. I'd suggest, each one of these is a separate column.

obernardovieira (approved):

PR #625, opened 2023-05-23
refactor exchange providers

### LINKED ISSUE #621: refactor backend exchange registry

obernardovieira:
This PR fixes #621

## Changes

adds new fields needed on the exchange page on the wallet

## Tests

no new tests

### DIFF
--- a/packages/core/src/services/app/cashoutProvider.ts
+++ b/packages/core/src/services/app/cashoutProvider.ts
@@ -65,28 +65,35 @@ export default class CashoutProviderService {
                 merchants = await models.merchantRegistry.findAll();
             }
 
-            merchants = merchants.filter((merchant) => {
-                const merchantLocation = point([
-                    merchant.gps.longitude,
-                    merchant.gps.latitude,
-                ]);
-                return (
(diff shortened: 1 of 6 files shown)


packages/core/src/services/app/cashoutProvider.ts
```diff
@@ -67,20 +70,11 @@ export default class CashoutProviderService {
                     merchant.gps.longitude,
                     merchant.gps.latitude,
                 ]);
-                const dist = distance(userLocation, merchantLocation);
-                if (dist < parseInt(query.distance || '100', 10)) {
-                    merchant['distance'] = dist;
```
dev-jotape, line 72: Without this line, will the `sort` in the function return continue working?

obernardovieira: this distance was not being given back on the endpoint and it was also not needed on frontend, since the results should already be sorted. So no need

obernardovieira: but actually, it's not sorted, good point. Let me do that

PR #653, opened 2023-05-31
fix firebase initialization

dev-jotape:
no task.

## Changes
Instead create a config file when `release-task.sh` runs, the firebase config JSON is recovered on `initPushNotificationService` (app.ts).


obernardovieira (approved):
smart ... 👀

PR #655, opened 2023-06-01
reduce check emails

### LINKED ISSUE #549: [LaE] notify admin level when funds are low
A given level can have X limit tokens to be disbursed. Although, not all tokens need to be sent at once. So the system must notify the admin level when funds are running low or reaching an end. Notify by email. Admin maybe should be a new table, with name and email and a wallet address. This will allow an account to sponsor one or multiple levels and keep tracking of it's progress.

obernardovieira:
This PR fixes #549

## Changes

Currently, too many emails are sent regarding Learn and Earn and hot wallets balance. This PR fixes it. Also updates associations.

## Tests

no new tests

### DIFF
--- a/packages/core/src/database/models/learnAndEarn/learnAndEarnPayment.ts
+++ b/packages/core/src/database/models/learnAndEarn/learnAndEarnPayment.ts
@@ -4,6 +4,7 @@ import {
     LearnAndEarnPayment,
     LearnAndEarnPaymentCreation,
 } from '../../../interfaces/learnAndEarn/learnAndEarnPayment';
+import { DbModels } from '../../../database/db';
 
 export class LearnAndEarnPaymentModel extends Model<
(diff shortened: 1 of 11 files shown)


packages/core/src/database/models/learnAndEarn/learnAndEarnPayment.ts
```diff
@@ -42,7 +44,7 @@ export function initializeLearnAndEarnPayment(
             levelId: {
                 type: DataTypes.INTEGER,
                 references: {
-                    model: 'learn_and_earn_lesson',
+                    model: learnAndEarnLesson,
```
dev-jotape, line 47: actually, this column should reference `learnAndEarnLevel`, not lesson (it was wrong before too)

dev-jotape (changes requested):

PR #676, opened 2023-06-12
[CHORE] remove unused envs

obernardovieira:
Remove some unused old envs


packages/api/src/middlewares/index.ts
```diff
@@ -81,27 +81,7 @@ export function optionalAuthentication(req: RequestWithUser, res: Response, next
...
-        }
-        const admin = _admin as { key: string };
-        //
-        if (config.adminKey !== admin.key) {
-            res.sendStatus(403);
-            return;
-        }
-        next(); // pass the execution off to whatever request the client intended
-    });
+    next();
```
dev-jotape, line 84: with this the push-notification endpoint will be open, is that right?

dev-jotape (changes requested):

dev-jotape (approved):

PR #681, opened 2023-06-15
lambda authorizer

### LINKED ISSUE #564: [MicroCredit] allow any user to sign up
See designs https://www.figma.com/file/nvpYzCMpk9fHJhGTeb9beY/Microcredit?node-id=7520%3A206641&t=YeG10TFnDjKjSZUk-1

TODO
- allow uploading files to a different bucket. Files should be private and protected (different than all buckets we've used until now)
- save survey on each step
- make survey modular. Survey can change. Questions  are on prismic only.

dev-jotape:
This PR fixes #564
## Changes
Create a lambda to be used as an authorizer on API Gateway.
This authorizer will validate a JWT Token before triggering another lambda.

### DIFF
--- a/services/authorizer/.gitignore
+++ b/services/authorizer/.gitignore
@@ -1,0 +1,5 @@
+# package directories
+node_modules
+
+# Serverless directories
+.serverless
--- a/services/authorizer/.mocharc.json
+++ b/services/authorizer/.mocharc.json
@@ -1,0 +1,9 @@
+{
+    "extension": ["ts"],
+    "spec": "tests/**/*.test.ts",
+    "require": ["ts-node/register", "tsconfig-paths/register"],
+    "delay": false,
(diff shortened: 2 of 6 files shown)


obernardovieira (approved):
Don't forget to document about how is this JWT used.
I have some questions, as examples.
How are the requests going to be done now? Same as before but requiring JWT? How's the JWT generated? How does that work with the current JWT we have?

dev-jotape:
@obernardovieira I updated the README.md adding a paragraph to explain the object of this lambda.
About your questions:
- The request will be the same, just need to add the JWT Token on the "Authorization" headers.
- The token can be fixed, generated using the same "secret key" used in this lambda (different from the token used on the API)

obernardovieira:
Thanks. Looks good. Can be merged :)

PR #688, opened 2023-06-20
clean community table

### LINKED ISSUE #439: clean community table
drop
* publicId
* descriptionEn
* coverImage
* coverMediaId

dev-jotape:
This PR fixes #439 

## Changes
Remove unused columns from community table:
- publicId
- descriptionEn
- coverImage
- coverMediaId


packages/core/src/database/migrations/z1687287269-update-community.js
```diff
@@ -0,0 +1,21 @@
...
+        await queryInterface.removeConstraint('ubi_request_change_params', 'ubi_request_change_params_communityId_fkey');
+        
+        await queryInterface.removeColumn('community', 'publicId');
+        await queryInterface.removeColumn('community', 'descriptionEn');
+        await queryInterface.removeColumn('community', 'coverImage');
+        await queryInterface.removeColumn('community', 'coverMediaId');
+    },
+
+    down(queryInterface, Sequelize) {},
+};
```
obernardovieira, line 21: remember that we have [preboot](https://devcenter.heroku.com/articles/preboot) enabled. If you run this while the current backend is using it, it's going to break. One possible way is to allow it to run only after X timestamp (like, in 1 month).

obernardovieira (approved):
Looks great. Just be careful with that migration that removes the columns.

packages/core/src/database/migrations/z1687287269-update-community.js
dev-jotape: I forgot this detail...
I moved the migration to the "future_migrations" to run later, I think it's easier right?

obernardovieira: No relly, because we will forget I think. There are a few more there that need to be moved. But lets merge it now

PR #702, opened 2023-06-29
fix community ubi demographics

### LINKED ISSUE #605: [UBI] rebuild demographics calculation
Currently, the demographics calculation for UBI is based on the "beneficiary" table. That table has really old data. Needs to be removed and beneficiaries need to be fetched from the subgraph.

dev-jotape:
This PR fixes #605 

## Changes
changed the `community-metrics` lambda to calculate the community ubi demographics getting the beneficiaries from TheGraph.
function added on the `calcuateCommunitiesMetrics` file to reuse the beneficiary's addresses already obtained to calculate the metrics for each community.


obernardovieira (approved):
Nice one reusing the query.

PR #773, opened 2023-07-26
fix count on get notification

dev-jotape:
no task.

## Changes
Fix `count` on get notification endpoint

### DIFF
--- a/packages/core/src/services/app/user/index.ts
+++ b/packages/core/src/services/app/user/index.ts
@@ -360,15 +360,18 @@ export default class UserService {
                 isWallet
             };
         }
+        if (unreadOnly !== undefined) {
+            where = {
+                ...where,
+                read: !unreadOnly
+            };
+        }
 
         const notifications: {
             rows: AppNotification[];
             count: number;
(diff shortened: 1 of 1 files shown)


packages/core/src/services/app/user/index.ts
```diff
@@ -361,40 +361,22 @@ export default class UserService {
...
-                order: [['createdAt', 'DESC']]
-            });
-        }
+        const notifications: {
+            rows: AppNotification[];
+            count: number;
+        } = await models.appNotification.findAndCountAll({
+            where: {
+                ...where,
+                ...(unreadOnly !== undefined ? { read: !unreadOnly } : {})
```
obernardovieira, line 370: I suggest you add this on the lines above, doing a validation like it is for `if (isWebApp !== undefined)` and `if (isWallet !== undefined)`, leaving this much clean with a simple `where`

dev-jotape: much better!

obernardovieira (approved):

PR #795, opened 2023-08-02
[feature] create update-borrowers lambda

dev-jotape:
no task.

## Changes
Create `update-borrowers` lambda, to every 30 minutes, get the borrowers updated on subgraph, and save on `subgraph_microcredit_borrowers`.


packages/core/src/database/migrations/z1690911789-create-subgraphMicrocreditBorrowers.js
```diff
@@ -0,0 +1,57 @@
...
+                allowNull: true, 
+            },
+            createdAt: {
+                type: Sequelize.DATE,
+                allowNull: false
+            },
+            updatedAt: {
+                type: Sequelize.DATE,
+                allowNull: false
+            },
```
obernardovieira, line 51: I would not include timestamps on this table

packages/core/src/database/migrations/z1690911789-create-subgraphMicrocreditBorrowers.js
```diff
@@ -0,0 +1,57 @@
...
+            userId: {
+                type: Sequelize.INTEGER,
+                allowNull: false,
+                references: {
+                    model: 'app_user',
+                    key: 'id',
+                },
+                onDelete: 'CASCADE'
+            },
```
obernardovieira, line 19: Instead, I suggest having an `address` field, that is actually the primary key and references to `app_user.address`. Any id as the primary key on this table is useless. Having the `userId` requires fetching user Ids from the address when updating, and I think it won't improve any aspect of the queries being done later on. But I might be wrong. What do you think?

dev-jotape: I thought a lot about using the `userId` or `address`. In the process to populate this table is much easier to use the `address`, but to list borrowers is better to use `userId` because the `microCreditBorrowers` table (and others) uses the user id as the reference of the user.
So following the pattern that we used for the other tables/associations I preferred the `userId` to facilitate on list borrowers.

packages/core/src/database/migrations/z1690911789-create-subgraphMicrocreditBorrowers.js
dev-jotape: I'm thinking to use the `updatedAt` to know the last updated date and filter the borrowers on subgraph from this date

obernardovieira: You should always get borrowers from the subgraph that have been updated in the last X minutes, regardless of what `updatedAt` you have on the database.
And you don't need to know when they were last updated. You can just assume they were updated because if the job fails, we will know.
Maybe I'm missing something. At least the `createdAt` it's definitely not needed.

services/microcredit/src/borrowers.ts
```diff
@@ -0,0 +1,55 @@
...
+    utils.Logger.info('Updating borrowers...');
+    const { subgraphMicroCreditBorrowers, appUser } = database.models;
+    const t = await database.sequelize.transaction();
+    try {
+        const lastUpdate = await subgraphMicroCreditBorrowers.findOne({
+            attributes: ['updatedAt'],
+            order: [['updatedAt', 'DESC']]
+        });
+
+        const entityLastUpdated = lastUpdate ? (lastUpdate.updatedAt.getTime() / 1000) | 0 : undefined;
```
obernardovieira, line 14: Ahh, I see what you are doing. But doing it this way will basically always fetch all users. The goal is to fetch only the ones updated in the last X minutes. One possible way is to save the timestamp of the last sync (update). It doesn't matter when each one of them was last updated. What matters is when was the last time that data was synced (fetched and updated successfully).

services/microcredit/src/borrowers.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        const borrowers = await subgraph.queries.microcredit.getBorrowers({ entityLastUpdated, limit: 1000 });
```
obernardovieira, line 16: what if there are more than a 1000 updates?

obernardovieira (approved):

PR #849, opened 2023-09-01
[feature] L&E notification

### LINKED ISSUE #745: [PushNotifications] Learn&Earn available courses and lessons
Included in this [PRD](https://docs.google.com/document/d/1b296B8PKMsUyB4UeU2Bf0ggcqh5eVKNWafmoK45ecF4/edit).

Once a week, notify beneficiaries about the Learn and Earn. If they have started any level but did not finish, notify them about that instead.

dev-jotape:
This PR fixes #745

## Changes
Update L&E webhook to save the `availableAt`.
New lambda to send notifications about L&E (new course, and remember a user to complete a course).


services/learn-and-earn/src/notification.ts
```diff
@@ -0,0 +1,105 @@
...
+        where: {
+            userId: {
+                [Op.in]: userIds,
+            },
+            type: interfaces.app.appNotification.NotificationType.LEARN_AND_EARN_FINISH_LEVEL,
+            createdAt: {
+                [Op.gt]: oneWeekAgo,
+            }
+        },
+    });
```
obernardovieira, line 74: We need to reduce this to one query. Or this `[Op.in]: userIds` will be massive at some point. I mean, it will fails, because there are certainly limits.

services/learn-and-earn/src/notification.ts
```diff
@@ -0,0 +1,105 @@
...
+import admin from 'firebase-admin';
+
+export async function availableCourses() {
+    const { learnAndEarnPrismicLevel, learnAndEarnUserLevel, appUser } = database.models;
+    const yesterday = new Date();
+    yesterday.setDate(yesterday.getDate() - 1);
+    yesterday.setHours(0, 0, 0, 0);
+
+    const newCourses = await learnAndEarnPrismicLevel.findAll({
+        where: {
```
obernardovieira, line 12: query only what is needed
```suggestion
        attributes: ['language'],
        where: {
```

obernardovieira (changes requested):

services/learn-and-earn/src/notification.ts
dev-jotape: Ok, I added a piece of SQL to reduce this in one query. What do you think?

obernardovieira: The problem with "in" and "not in" is that, if the select gives 50k results, then you'll have an insanely long query. Which, it's my problem here. I don't think such a long query will run. It'll just throw an error.

Because of that, I had in mind that, it is less performant. And after some research, it seems like it is
https://www.sqlshack.com/t-sql-commands-performance-comparison-not-vs-not-exists-vs-left-join-vs-except/
https://explainextended.com/2009/09/16/not-in-vs-not-exists-vs-left-join-is-null-postgresql/

I'll leave you here the answer from chatgpt, which is what I was thinking.

> To find users who started a level one week or more ago but were not notified less than a week ago, you can use a SQL query that joins the `user`, `user_notification`, and `user_learn_and_earn` tables and filters the results based on the conditions you mentioned. Here's a possible SQL query to achieve this:

```sql
SELECT u.user_id, u.username
FROM users u
JOIN user_learn_and_earn_level l ON u.user_id = l.user_id
LEFT JOIN (
    SELECT user_id, MAX(notification_date) AS last_notification_date
    FROM user_notification
    WHERE notification_type = 'level_started'
    GROUP BY user_id
) ln ON u.user_id = ln.user_id
WHERE
    l.status = 'started'
    AND (CURRENT_TIMESTAMP - l.start_date) >= INTERVAL '7 days' -- Started one week or more ago
    AND (ln.last_notification_date IS NULL OR (CURRENT_TIMESTAMP - ln.last_notification_date) >= INTERVAL '7 days')
```

> This query does the following:
> 1. It selects the `user_id` and `username` from the `users` table.
> 2. It joins the `user_learn_and_earn_level` table to get information about users who started a level. We assume there's a `start_date` column in this table.
> 3. It left joins a subquery that calculates the last notification date for each user who received a "level_started" notification.
> 4. It applies the following conditions:
>    - Users must have a status of 'started' in the `user_learn_and_earn_level` table.
>    - Users must have started their level one week or more ago.
>    - Users should either have no previous notifications or have received their last notification one week or more ago.
> 
> This query will return a list of users who meet these criteria and need to be notified. You can adjust the column names and table names to match your actual database schema.

If you want to go even further, you can try to compare those queries with [EXPLAIN](https://www.cybertec-postgresql.com/en/how-to-interpret-postgresql-explain-analyze-output/).

obernardovieira (approved):

PR #862, opened 2023-09-05
[feature] - cache learn and earn

dev-jotape:
This PR fixes #gh-issue-number

## Changes
Cache GET endpoints from L&E and clean when the user interacts with the lessons or when the webhook is trigged


obernardovieira:
refactored to clean cache on services rather than endpoints. Also, cache using userid instead of ip requesting (it breaks if multiple users use the same device + we don't want to use ip's)

obernardovieira (approved):

PR #906, opened 2023-09-19
[bugfix] - Convert data to string before setting cache

dev-jotape:
This PR fixes #gh-issue-number

## Changes
Redis was crashing on trying to convert arrays to strings to save on the cache. So, we are first converting the data to string before calling the `redis.set()`.
Additionally, we add the library `typia` to start some experiments. First we are using the `typia.stringify` and `typia.isParse` on redis middleware.


packages/api/package.json
```diff
@@ -41,7 +42,10 @@
         "express-status-monitor": "1.3.4",
         "mocha": "10.2.0",
         "nodemon": "3.0.1",
-        "prettier": "3.0.0"
+        "prettier": "3.0.0",
+        "ts-node": "10.9.1",
+        "ts-patch": "3.0.2",
+        "typescript": "5.2.2"
```
obernardovieira, line 48: hmm, I think you don't need typescript here since you have it on dependencies

obernardovieira (approved):

packages/api/package.json
dev-jotape: you are right!

PR #914, opened 2023-09-20
[feature] index community data from subgraph

### LINKED ISSUE #905: index community data from subgraph, the same way as the borrowers
it will fix this https://appsignal.com/impactmarket/sites/63c5932e83eb6790d1b8c0f4/exceptions/incidents/85

dev-jotape:
This PR fixes #905 

## Changes
Index community data from subgraph.
- new table to save the community state `subgraph_community`
- Microservice to update community data every day (communities that had activities)
- change the community list


services/community-metrics/src/updateCommunities.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        await Promise.all(communities.map(community => {
+            return subgraphCommunity
+                .findOne({ where: { communityAddress: getAddress(community.id) } })
+                .then(subgraphCommunityModel => {
+                    community['communityAddress'] = getAddress(community.id);
+                    delete community.id;
+
+                    if(subgraphCommunityModel)
+                        return subgraphCommunityModel.update(community, { transaction: t });
```
obernardovieira, line 31: please, don't skip the curly brackets (easier to read)
```suggestion
                    if(subgraphCommunityModel) {
                        return subgraphCommunityModel.update(community, { transaction: t });
                    }
```

services/community-metrics/src/updateCommunities.ts
```diff
@@ -0,0 +1,43 @@
...
+        const communities = await subgraph.queries.community.communityEntities(
+            `first: 1000,
+            ${entityLastUpdated ? `where: { lastActivity_gt: ${entityLastUpdated} }` : ''}`,
+            `id, estimatedFunds, beneficiaries, claims, claimed, beneficiaries, removedBeneficiaries, contributed, contributors, managers, baseInterval, state`
+        );
+
+        await Promise.all(communities.map(community => {
+            return subgraphCommunity
+                .findOne({ where: { communityAddress: getAddress(community.id) } })
+                .then(subgraphCommunityModel => {
```
obernardovieira, line 26: maybe you can call it `entity`, rather than `subgraphCommunityModel` since it might be confused with the database model

obernardovieira (approved):
wow... the size that it reduced on that list 🤯 

Left two suggestions and want to ask something.
And I have a request. Could you please rename `SubgraphCommunity` to `SubgraphUBICommunity`. I also want to rename `community` to `UBICommunity` , but not now. The reason is that, that way, we have all the tables organized when we look at the database. It's easier to read (and query, we would always know it start with UBI or MicroCredit or Story). And we keep some standard. So, we did that for, Story, Learn and Earn and MicroCredit, and everything else is App<Something> and it would be awesome to at some point refactor past tables, entity, etc, but while we don't, lets keep a pattern for new ones :) 

Then feel free to merge.

services/community-metrics/src/updateCommunities.ts
dev-jotape: ok, done!

PR #950, opened 2023-10-06
[feature] subgraph beneficiary

### LINKED ISSUE #902: index beneficiary data from subgraph, the same way as the borrowers
Like the borrowers https://github.com/impactMarket/backend/pull/795

Queries lists (like, list beneficiaries) should then use the data indexed instead of subgraph queries.

dev-jotape:
This PR fixes #902 

## Changes
Index beneficiary state on `subgraph_ubi_beneficiary`


packages/core/src/interfaces/ubi/subgraphUBIBeneficiary.ts
```diff
@@ -0,0 +1,25 @@
...
+    claims: number;
+    claimed: number;
+    lastClaimAt: number;
+    preLastClaimAt: number;
+    since: number;
+    state: number;
+
+    // timestamps
+    updatedAt: Date;
+}
```
obernardovieira, line 14: If we look at loan managers design, you will see that we only need `claimed`, `since`, and `state`. So all the other fields you can remove them.

One important note although. When a beneficiary is added or removed, this registry needs to be updated, otherwise it will not be updated on the frontend, resulting in bug reports.

obernardovieira: Thinking about this once again, maybe the `chainSubscriber` needs an `upsert` into this table both for BeneficiaryAdded and Beneficiary removed, only to update the state.

services/community-metrics/src/updateBeneficiaries.ts
```diff
@@ -0,0 +1,99 @@
...
+                    beneficiary['communityAddress'] = getAddress(beneficiary.community.id);
+                    beneficiary['userAddress'] = getAddress(beneficiary.address);
+    
+                    if(model) {
+                        return subgraphUBIBeneficiary.update(beneficiary as any, {
+                            where: { id: model.id },
+                            transaction: t,
+                        }) as any;
+                    }
+                    return subgraphUBIBeneficiary.create(beneficiary as any, { transaction: t });
```
obernardovieira, line 79: would it make sense to clear the beneficiaries' endpoints cache somewhere here?

obernardovieira (changes requested):

obernardovieira:
btw, I've recently watched the video here https://substreams.streamingfast.io/#where-to-start. It might bring some ideas for future improvements.

packages/core/src/interfaces/ubi/subgraphUBIBeneficiary.ts
dev-jotape: done!

services/community-metrics/src/updateBeneficiaries.ts
dev-jotape: Yes, totally sense

packages/worker/src/chainSubscribers.ts
```diff
@@ -335,12 +335,22 @@ class ChainSubscribers {
                 const userAddress = parsedLog.args[1];
 
                 if (community) {
+                    // add subgraph beneficiary
+                    await database.models.subgraphUBIBeneficiary.create({
```
obernardovieira, line 339: what if the beneficiary is removed from one community and added to another one?

services/community-metrics/src/updateBeneficiaries.ts
```diff
@@ -63,17 +63,20 @@ async function updateBeneficiariesState(searchParam: number | string,  t: Transa
...
-                        return subgraphUBIBeneficiary.update(beneficiary as any, {
+                        await subgraphUBIBeneficiary.update(beneficiary as any, {
                             where: { id: model.id },
                             transaction: t,
                         }) as any;
                     }
-                    return subgraphUBIBeneficiary.create(beneficiary as any, { transaction: t });
+                    await subgraphUBIBeneficiary.create(beneficiary as any, { transaction: t });
+
+                    utils.cache.cleanBeneficiaryCache(beneficiary.community.id);
```
obernardovieira, line 78: The `beneficiaries` come from the subgraph and `community.id` in this case is the address in lower case, not the actual community id. This method requires the id (number), right?

obernardovieira (changes requested):

packages/worker/src/chainSubscribers.ts
dev-jotape: the `userAddress` is not a unique key on this table. So, in this case will have 2 registries about the same user, but one with the state 0 and other state 1 - removed.

services/community-metrics/src/updateBeneficiaries.ts
dev-jotape: yes, you are right.
I made a change to clean the cache at the end of the process (getting the community ID)

packages/worker/src/chainSubscribers.ts
obernardovieira: ahh right, because you need to keep the old one. Thanks

obernardovieira (approved):

obernardovieira:
@dev-jotape approved. Just need to fix the conflicts and can be merged.

PR #1019, opened 2023-11-10
[feature] update current debit

### LINKED ISSUE #1012: [Microcredit] Periodically update the debt of borrowers when they did not repay for more that 3 weeks
To update, override the `currentDebt` in `subgraph_microcredit_borrowers` with the correct value from the smart contract https://github.com/impactMarket/smart-contracts-dao/blob/main/contracts/microcredit/MicrocreditImplementation.sol#L191

dev-jotape:
This PR fixes #1012 

## Changes
Create lambda to update current debit


services/microcredit/src/borrowers.ts
```diff
@@ -63,3 +66,62 @@ export async function updateBorrowers(): Promise<void> {
...
+        threeWeeksAgo.setDate(threeWeeksAgo.getDate() - 21);
+
+        const borrowers = await subgraphMicroCreditBorrowers.findAll({
+            include: [{
+                model: appUser,
+                as: 'user',
+            }],
+            where: {
+                lastRepayment: {
+                    [Op.lt]: threeWeeksAgo.getTime() / 1000 | 0
```
obernardovieira, line 87: Just thought of this now. It will start to update every day then, but it shouldn't. Do we have an "updatedAt" on this table. If so, can we update the debt, only if the lastRepayment and updatedAt are both older than 3 weeks?

dev-jotape: Got it!
I changed to check the updatedAt and lastRepayment

obernardovieira (approved):

PR #1058, opened 2023-11-24
[refactor] endpoint params validator

dev-jotape:
## Changes
Refactor endpoint validators to validate the request params.

### DIFF
--- a/packages/api/src/controllers/v2/community/details.ts
+++ b/packages/api/src/controllers/v2/community/details.ts
@@ -1,6 +1,6 @@
 import { Request, Response } from 'express';
+import { config, services } from '@impactmarket/core';
 import { getAddress } from '@ethersproject/address';
-import { services } from '@impactmarket/core';
 
 import { RequestWithUser } from '../../../middlewares/core';
 import { standardResponse } from '../../../utils/api';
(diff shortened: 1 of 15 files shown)


packages/api/src/controllers/v2/community/details.ts
```diff
@@ -16,30 +16,18 @@ class CommunityController {
     getManagers = (req: RequestWithUser, res: Response) => {
```
obernardovieira, line 16: We can do something like https://github.com/impactMarket/backend/blob/main/packages/api/src/controllers/v2/microcredit/list.ts#L22, so we avoid the casts like `orderBy as string` or  `lastActivity_lt as string` below.

packages/api/src/controllers/v2/community/details.ts
```diff
@@ -57,48 +45,31 @@ class CommunityController {
...
-        }
+        const { state, offset, limit, orderBy } = req.query;
         this.detailsService
             .listBeneficiaries(
                 req.user.address,
                 parseInt(req.params.id, 10),
-                parseInt(offset, 10),
-                parseInt(limit, 10),
+                offset ? parseInt(offset as string, 10) : config.defaultOffset,
+                limit ? parseInt(limit as string, 10) : config.defaultLimit,
```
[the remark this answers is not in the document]
obernardovieira: Also, be careful with where are the config is fetched from. I's being fetched from core, but in this case, it should be from the api

obernardovieira (changes requested):

packages/api/src/controllers/v2/community/details.ts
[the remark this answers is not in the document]
dev-jotape: Nice! I changed to follow this pattern

packages/api/src/controllers/v2/community/details.ts
dev-jotape: done!

packages/api/src/validators/community.ts
```diff
@@ -1,7 +1,38 @@
 import { Joi, celebrate } from 'celebrate';
 
+import { ContainerTypes, ValidatedRequestSchema } from '~utils/queryValidator';
+import { config } from '@impactmarket/core';
```
obernardovieira, line 4: The config here is used for `defaultLimit` and `defaultOffset`. This are api-specific, so avoid using the core config.

packages/api/src/validators/story.ts
```diff
@@ -1,18 +1,28 @@
 import { Joi, celebrate } from 'celebrate';
 
-import { createValidator } from '../utils/queryValidator';
+import { ContainerTypes, ValidatedRequestSchema, createValidator } from '../utils/queryValidator';
+import { config } from '@impactmarket/core';
```
obernardovieira, line 4: Same about the config.

